PR #12, opened 2019-02-28
We prefer mplex over yamux. 

gnunicorn:
Fixes talking to go-ipfs


dvc94ch:
This breaks talking from a rust-ipfs node to another rust-ipfs node. This behaviour can also be seen with the libp2p chat application.

dvc94ch:
I opened two issues #23 #24 and reverted this commit for now.

PR #30, opened 2019-03-22
Extending unixfs

gnunicorn:
to read files, recursively, using protobuf and understanding the underlying formats. Compatible with the default hello world block at `QmR7tiySn6vFHcEjBeZNtYGAFh735PJHfEMdVEycj9jAPy`

ToDo:

 - [ ] add protobuf generation to script
 - [ ] add tests to proof this works as proclaimed


dvc94ch:
Why not use the existing `File` abstraction? Also the `IpldDag` is supposed to abstract over fetching blocks and adding them to a dag. So following links, or resolving a `{cid}/styles/style.css` should be a single operation for the gateway. You can turn that into an IpfsPath and then the `IpldDag` will fetch the missing blocks.

dvc94ch:
ping @gnunicorn :)

gnunicorn:
@dvc94ch sorry, was off-duty for a few weeks.

I did have a changeset attempting to use `File`, I recall. It just didn't exist yet to this extend, when I first attempting to implement the warp-based HTTP serving gateway. However, I am not sure I'd want to follow the current approach taken here (or with the `File`-Interface to be honest) because I don't quite get the inner-relationship of all these formats, which ones are deprecated but still in use and how they were even meant to be used.

dvc94ch:
I agree with you. The specs/docs/go/js don't match at all. `go-ipld-prime` seems (even with some issues not worked out yet) to be the way `ipld` is going. `unixfsv2` only has a js implementation currently, and it seems that that will eventually fit onto `go-ipld-prime`. Incidentally our `ipld` implementation is most similar to `go-ipld-prime` so I think we're on the right track there, but we should probably factor it out into it's own crate (go-ipld-prime contains most notably a schema dsl and validation logic that is still missing). When it comes to how unixfs interacts with `ipld` it's not so clear yet how that will end up.

So how do you suggest we move forward?

dvc94ch:
I'm making progress on a standalone redesigned ipld library. The current design has some limitations.

PR #33, opened 2019-11-22
Get it compiling again

koivunej:
Basically just back-off from the dependency to [now closed rust-libp2p pr](https://github.com/libp2p/rust-libp2p/pull/968) and guess a rust toolchain that works. I'm publishing this as a draft since there has been lots of interest in this repo and maybe this is an easier path forward than to jump immediatedly to latest rustc and deps. What are your plans for rust-ipfs in the near future?

All the tests pass but communication with `go-ipfs` 0.4.22 does not seem to work (the `examples/ipfs_bitswap_test`, connecting directly through bootstrap config). Not sure why yet.


dvc94ch:
Communication with go-ipfs is not expected to work, since the rust/go/js mdns implementations where mutually incompatible. Using the dht didn't work either. I never got around to figuring out what the problem is. From the changes you've made I'd expect rust-ipfs to be compatible only with itself.

koivunej:
Thanks for your response! Perhaps I should just then drop the draft flag and continue from here investigating what it would require to make the communication work? I think I looked that the mdns work is still ongoing for js/go.

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2019-02-26
```
koivunej: This should match the travis version I didn't think to check...

dvc94ch:
Awesome, thanks!

Issue #36, opened 2019-12-05
Drop nightly requirement


koivunej:
Moving to stable would allow us to keep up to date easier, and allow tools like [`skeptic`](https://crates.io/crates/skeptic) and compile the README.md example automatically. Currently there is only single feature being used: `try_trait` which could be replaced [`void`](https://crates.io/crates/void) which is already part of the dependency tree.

austinsheep:
For replacing `try_trait` with `void`, would the error cases for the `impl TryInto/TryFrom`'s instead call `unreachable!()`?

koivunej:
Here was a comment asking about this the `void` hint. Could be that I've misunderstood the current use of `try_trait` and the suggestion to replace it with a `void::Void` is wrong. Could be that a simple error enum or empty struct would work here better.

dvc94ch:
I agree, now that async await is stable, we should drop the nightly requirement

koivunej:
Implemented in #39

PR #39, opened 2019-12-17
Migrating to stable: Replaced NoneError with TryError

austinsheep:
Sorry for removing my original comment to the relevant issue #36 where I was asking about using `void::Void` to remove the `try_trait` feature.

In this PR I've added the unit struct, `TryError`, to `src/error.rs` to replace all cases of `NoneError`.

All tests appear to succeed when executing `cargo test` from stable, but I was unsure how to test the example files.

### DIFF
--- a/src/ipld/formats/pb/mod.rs
+++ b/src/ipld/formats/pb/mod.rs
@@ -1,5 +1,5 @@
 use crate::block::Cid;
-use crate::error::Error;
+use crate::error::{Error, TryError};
 use crate::ipld::Ipld;
 use crate::path::PathRoot;
 use cid::Prefix;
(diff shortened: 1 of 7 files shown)


src/ipld/formats/pb/mod.rs
```diff
@@ -96,41 +96,56 @@ impl Into<Ipld> for PbLink {
...
 
 impl TryFrom<Ipld> for PbNode {
-    type Error = std::option::NoneError;
+    type Error = TryError;
 
     fn try_from(ipld: Ipld) -> Result<PbNode, Self::Error> {
         match ipld {
             Ipld::Object(mut map) => {
-                let links: Vec<Ipld> = map.remove("Links")?.try_into()?;
+                let links: Vec<Ipld> = match map.remove("Links") {
```
dvc94ch: wouldn't `.map_or(TryError)` be nicer than using `match` everywhere?

src/ipld/formats/pb/mod.rs
```diff
@@ -96,41 +96,56 @@ impl Into<Ipld> for PbLink {
...
+                    Some(links) => links.try_into()?,
+                    None => return Err(TryError)
+                };
                 let links: Vec<PbLink> = links.into_iter()
                     .map(|link| link.try_into()).collect::<Result<_, Self::Error>>()?;
-                let data: Vec<u8> = map.remove("Data")?.try_into()?;
+                let data: Vec<u8> = match map.remove("Data") {
```
dvc94ch: same here

src/ipld/formats/pb/mod.rs
```diff
@@ -96,41 +96,56 @@ impl Into<Ipld> for PbLink {
...
-    type Error = std::option::NoneError;
+    type Error = TryError;
 
     fn try_from(ipld: Ipld) -> Result<PbLink, Self::Error> {
         match ipld {
             Ipld::Object(mut map) => {
-                let cid: PathRoot = map.remove("Hash")?.try_into()?;
-                let name: String = map.remove("Name")?.try_into()?;
-                let size: u64 = map.remove("Tsize")?.try_into()?;
+                let cid: PathRoot = match map.remove("Hash") {
```
dvc94ch: and here

src/ipld/formats/pb/mod.rs
austinsheep: The `.unwrap_or()` function doesn't return the `try_from` function, it tries to set an alternative `Ipld` value.

When I try to use `.unwrap_or(TryError)` here I get:
`src/ipld/formats/pb/mod.rs|104 col 70 error 308| mismatched types: expected enum 'ipld::ipld::Ipld', found struct 'error::TryError'`

koivunej:
The chaining looks good, I'll go ahead and merge these!

PR #43, opened 2020-01-08
Switch to rust-libp2p version with stable futures

rklaehn:
Update rust-libp2p dep to libp2p v0.14.0-alpha.1 (324f0dc326cfdbc2d40d1e134a5734513d809b4c)

and make it compile. Not really sure if everything works, but I think this is a good start.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ tokio = { version = "0.1.21" }
 tokio-fs = { version = "0.1" }
 xdg = "*"
 async-trait = "*"
+async-std = { version = "1.3.0", features = [ "unstable", "std" ] }
 
 rocksdb = { version = "*", optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ env_logger = "*"
 failure = "*"
 fnv = "*"
 futures = { version = "0.3", features = [ "compat", "io-compat", "async-await" ] }
-libp2p = { version = "0.13.1", git = "https://github.com/libp2p/rust-libp2p" }
+libp2p = { git = "https://github.com/libp2p/rust-libp2p", rev = "324f0dc326cfdbc2d40d1e134a5734513d809b4c" }
 log = "*"
 multibase = "*"
 multihash = "*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ env_logger = "*"
 failure = "*"
 fnv = "*"
 futures = { version = "0.3", features = [ "compat", "io-compat", "async-await" ] }
-libp2p = { git = "https://github.com/libp2p/rust-libp2p", rev = "324f0dc326cfdbc2d40d1e134a5734513d809b4c" }
+libp2p = "0.14.0-alpha.1"
 log = "*"
 multibase = "*"
 multihash = "*"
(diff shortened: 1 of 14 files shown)


dvc94ch:
Nice! Thanks for contributing. I saw on twitter you were planning on starting from scratch, glad you changed your mind. Looks good so far.

rklaehn:
I thought the switch to the new futures would be more painful, being more of a rewrite. But the change turned out to be pretty limited in scope.

Agree about switching to an org, but let's postpone this discussion for some time. I am not sure if I will be able to contribute continuously. I will know more at the end of next week.

There are a few things that should help with a more fully featured rust ipfs:

- I implemented tags in serde cbor. Should be merged and available quite soon, as an optional feature: 
https://github.com/pyfisch/cbor/pull/172

- @vmx has a mini repo implementing IPLD based on the cbor tags support above:
https://github.com/vmx/rust-ipld-dag-cbor

Cargo.toml
```diff
@@ -17,7 +17,7 @@ env_logger = "*"
 failure = "*"
 fnv = "*"
 futures = { version = "0.3", features = [ "compat", "io-compat", "async-await" ] }
-libp2p = { version = "0.13.1", git = "https://github.com/libp2p/rust-libp2p" }
+libp2p = { git = "https://github.com/libp2p/rust-libp2p", rev = "324f0dc326cfdbc2d40d1e134a5734513d809b4c" }
```
koivunej: was there an issue using the 0.14.0-alpha1 or whatever it's called, I think they published it on crates.io already?

koivunej:
Hello @rklaehn and nice to see a PR from you :) I actually looked into doing this yesterday, got bogged down with figuring out if the both executors are an issue. I see you chose just to add `async_std` and not touch the tokio 0.1 parts which I think I found out is ok to do.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ tokio = { version = "0.1.21" }
 tokio-fs = { version = "0.1" }
 xdg = "*"
 async-trait = "*"
+async-std = { version = "1.3.0", features = [ "unstable", "std" ] }
```
koivunej: similar question here, any particular reason for picking 1.3 over 1.4 (or even "1"), can't see any breaking changes in their [gh releases](https://github.com/async-rs/async-std/releases)?

Cargo.toml
rklaehn: I did not see that. Will update it.

Cargo.toml
rklaehn: I think it was just the latest when I started playing with this...

rklaehn:
> Just adding this note here so that's written somewhere: policits of async-std or tokio aside, I looked into removal tokio 0.1 dependency as we'll likely need to do now that rust-libp2p is chose `async-std`. I think the only explicit dependency on comes from the filesystem operations (which likely have `async-std` counterparts) and the `domain` dependency which takes in tokio 0.1 (there is a threadpooled system resolver access from `async-std`, not sure if it allows TXT records). Alternative would be `trusted-dns-resolver`.

I basically just chose async-std since @tomaka chose it for rust-libp2p. But I would be really interested in a neutral comparison of the two. We have the same decision upcoming for our internal code...

tomaka:
> Just adding this note here so that's written somewhere: policits of async-std or tokio aside, I looked into removal tokio 0.1 dependency as we'll likely need to do now that rust-libp2p is chose async-std

The entire reason why we switched to async-std is that it works everywhere and that there's no need to be concerned about which runtime you're using.

rklaehn:
This is *very* WIP, but does this go in rougly the right direction?

https://github.com/libp2p/rust-libp2p/pull/1385

rklaehn:
@dvc94ch are you OK with merging this? I won't be able to do much work on this next week since I have a company event.

dvc94ch:
LGTM, thanks!

PR #46, opened 2020-02-06
Use prost! instead of protobuf

rklaehn:
To be consistent with what rust-libp2p does...

Now that we only have such a small number of protocols, the switch is still relatively simple...


rklaehn:
Well, it passes the tests which do some roundtripping, so I guess it should be OK.

dvc94ch:
We should setup ci

PR #49, opened 2020-02-10
Add new bitswap protocol from go-ipfs

rklaehn:
Also fix quite embarrassing bug in the proto serialization

### DIFF
--- a/src/bitswap/bitswap_pb.proto
+++ b/src/bitswap/bitswap_pb.proto
@@ -3,29 +3,42 @@ syntax = "proto3";
 package bitswap_pb;
 
 message Message {
+
   message Wantlist {
+    enum WantType {
+      Block = 0;
+      Have = 1;
+    }
+
     message Entry {
-      // the block cid (cidV0 in bitswap 1.0.0, cidV1 in bitswap 1.1.0)
-			bytes block = 1;
-      // the priority (normalized). default to 1
-			int32 priority = 2;
-      // whether this revokes an entry
-			bool cancel = 3;
+			bytes block = 1;		// the block cid (cidV0 in bitswap 1.0.0, cidV1 in bitswap 1.1.0)
+			int32 priority = 2;	// the priority (normalized). default to 1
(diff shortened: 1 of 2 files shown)


src/bitswap/ledger.rs
```diff
@@ -171,6 +171,9 @@ impl Into<Vec<u8>> for &Message<O> {
             payload.data = block.data().to_vec();
             proto.payload.push(payload);
         }
+        if !wantlist.entries.is_empty() {
+            proto.wantlist = Some(wantlist);
+        }
```
koivunej: Yeah this looks important, good find!

src/bitswap/bitswap_pb.proto
```diff
@@ -3,29 +3,42 @@ syntax = "proto3";
...
+    }
+
     message Entry {
-      // the block cid (cidV0 in bitswap 1.0.0, cidV1 in bitswap 1.1.0)
-			bytes block = 1;
-      // the priority (normalized). default to 1
-			int32 priority = 2;
-      // whether this revokes an entry
-			bool cancel = 3;
+			bytes block = 1;		// the block cid (cidV0 in bitswap 1.0.0, cidV1 in bitswap 1.1.0)
```
dvc94ch: looks like mixed tabs and spaces?

rklaehn:
I have just copied the proto file from the go-bitswap repo and removed some go specific extensions.

src/bitswap/ledger.rs
dvc94ch: looks like it was introduced in #46

src/bitswap/bitswap_pb.proto
rklaehn: yes, it is a basically verbatim copy of the file from go-ipfs. Did not bother to format it. Does it matter?

dvc94ch: I'd have formatted it but it's done now

Issue #42, opened 2019-12-27
rustfmt run and travis integration


koivunej:
Similar to how clippy was added, rustfmt could be integrated. I think the codebase is already quite rustfmt-ish so there shouldn't be huge changes.

Adding good first issue as there shouldn't be any issues here. Travis might not install the `rustfmt` component by default as it did not install `clippy` either. Before making the formatting changes make sure the `cargo fmt -- --check` or whatever CI command fails as there currently should be some formatting issues.

aphelionz:
This appears to be done, yeah? Can we close?

PR #53, opened 2020-02-23
Cleanup

dvc94ch:
Performed some cleanup and updated the dependencies.

The dns is currently broken due to tokio-timer only working on a tokio executor [0].
- https://github.com/rust-lang/futures-rs/issues/1285

The domain crate was updated to use the master branch however as a bug was fixed that causes a panic.
- https://github.com/NLnetLabs/domain/pull/14

It's a large diff because I ran rustfmt, but you can review the commits individually.


src/repo/mod.rs
```diff
@@ -133,35 +133,15 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
...
-                None => {
-                    if once {
-                        once = false;
-                        info!("getting block later: {}", cid);
-                    }
-                    Delay::new(Instant::now() + Duration::from_millis(5000)).compat().await.unwrap();
-                    continue
-                },
+                Some(block) => return Ok(block),
+                None => continue,
```
dvc94ch: I think this was broken in the async-std transition? I removed it as it's a hack and won't work without the tokio executor anymore

koivunej: I think this will spin busy without the hack, but yeah it'll need some real solution.

src/ipns/dns.rs
```diff
@@ -74,9 +76,10 @@ mod tests {
     use crate::tests::async_test;
 
     #[test]
+    #[ignore]
```
koivunej: Forgot this while jumping between commits, was expecting it to be removed: Why was this ignore added?

koivunej:
This will partly resolve #42 (the rustfmt part).

src/ipns/dns.rs
dvc94ch: Because domain uses tokio-timer to timeout dns queries, and they require a tokio executor, they timeout immediately, meaning that the test will fail with a timeout error. As a side note I've been working on hashgraph the past week and think that we could remove the entire ipns/dns stuff in favor of a hashgraph based solution in the not too distant future.

koivunej: Oh... Doesn't this mean that the dns side is broken with this PR?

Yeah I was thinking that the whole dependency would need to be switched for trust-dns.

Could you elaborate how hashgraph solution could replace ipns/dns?

dvc94ch: This PR makes the switch to the async-std executor, but dns was broken before due to the bugfix not being included

dvc94ch: so with hashgraph I'm working on a byzantine fault tolerant distributed key value store with a dynamic set of peers. If that turns out to work well, it could be ipns 2.0. The other place where I see it being useful in ipfs is in bitswap for addressing the problems I mentioned with it's incentive mechanism. We can discuss it further in a chat if you're interested.

src/repo/mod.rs
dvc94ch: So the repo should have a `HashMap<Cid, Vec<Context>>` or something like that.

dvc94ch: If we migrate to rust-ipld this stuff will change anyway. So the store trait should have a `poll_get` instead of an `async get` and the store should register the future context if the block is not in the store and block. The deamon needs to then send a notify event and the client needs to wake up all the registered contexts for the cid.

src/ipns/dns.rs
koivunej: Cool! I was just making sure I guessed the scale, as in "ipns 2.0" and "nothing compatible with current ipns" correct. Yeah, let's follow up with this on the chat.

koivunej:
I think we could move forward with merging this, while I am not too happy on the state of dns nor the removal of the delay hack with the store, this PR is still major win the formatting, dns panic fix and other cleanup.

RE: in my [earlier comment](https://github.com/ipfs-rust/rust-ipfs/pull/53#pullrequestreview-367035597):
> I'll open a roadmap issue later today for longer discussion

I failed to do that yesterday so I did it this morning, please find it at #59.

src/repo/mod.rs
koivunej: > `HashMap<Cid, Vec<Context>>`

I think something like this, with something representing the multihash to a list of Wakers could work. Need to check if the latest developments have brought some tools for notifying N tasks at that point.. Can't find from the new primitives in tokio helping here nor immediatedly see how this could be done. Using the receiver side of an one-shot channel wrapped inside a `Arc<Mutex<_>>` to get every task waiting first waiting for mutex, then waiting for a message in turns. Upon completion a sender could be dropped, which would release everyone waiting for messages. That sounds also like a hack :)

dvc94ch: I was thinking something like this?

```rust
for ctx in contexts {
    ctx.wake()
}
```

PR #55, opened 2020-02-26
docs: use the Rust IPFS logo

alanshaw:
I think this looks great 😁 @aphelionz


aphelionz:
Agreed. Thank you so much for the logo @alanshaw

aphelionz:
Looks like we're having issues with the camo.github proxy, but I saw that we're to expect some downtimes today so we'll keep an eye on it

PR #64, opened 2020-03-04
CONTRIBUTING.md

aphelionz:
Please everybody, share your input with regards to CONTRIBUTING.md best practices!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,6 @@ Welcome, and thank you for your interest in contributing to Rust IPFS. Issues an
 
 1. Keep the build time small
 2. Aim for high (but not absolute) code coverage in testing
-3. Keep the dependency count low (use [`cargo-deps`](https://github.com/m-cat/cargo-deps))
 
 ## Target Build
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ We welcome all forms of contribution. Please open issues and PRs for:
 
 ## Style
 
-- Git Commit Messages should lean towards Conventional Commits but will not be enforced
+- Git Commit Messages should lean towards [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) but will not be enforced
 - Rust code should conform to `rustfmt` and `clippy` before push, as the CI will catch errors there
 
 ## Security vuln disclosure - perhaps in SECURITY.md?
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,31 +1,37 @@
 # Contributing to Rust IPFS
 
-Welcome, and thank you for your interest in contributing to Rust IPFS. Issues and pull requests are encouraged.
+Welcome, and thank you for your interest in contributing to Rust IPFS. Issues and pull requests are encouraged. As we work through the efforts described in the approved IPFS Dev Grant, we're looking for help in the following
+
+* Implementing endpoints and features not covered in the dev grant (See the [README](./README.md#roadmap))
+* Tests and CI for existing functionality
+* Examples
+* Documentation
+
+## First Principles
+
+1. Keep the build time small
+2. Aim for high (but not absolute) code coverage in testing
+3. Keep the dependency count low (use [`cargo-deps`](https://github.com/m-cat/cargo-deps))
 
 ## Target Build
 
-Rust IPFS will always target the current _stable_ version of Rust that is released. Our CI/CD tests will reflect this.
-
-TODO: Instructions on installing the up-to-date stable toolchain
-TODO: First Principles
-
-## Roadmap
-
-- Link to roadmap issue or project?
+Rust IPFS will always target the current _stable_ version of Rust that is released. Our CI/CD tests will reflect this. See [instructions here on how to install the rust toolchain](https://doc.rust-lang.org/book/ch01-01-installation.html).
 
 ## Contributing
 
-Reporting Bugs
-Suggesting Enhancements
-Your First Code Contribution
-Pull Requests
+We welcome all forms of contribution. Please open issues and PRs for:
 
-## TODO: Styleguides
+- Reporting Bugs
+- Suggesting Enhancements
+- Adding new tests
+- Adding new functionality
+- Documentation-only udpates
 
-Git Commit Messages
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+
+## TODO: Styleguides
+
+Git Commit Messages
+JavaScript Styleguide
+CoffeeScript Styleguide
+Specs Styleguide
+Documentation Styleguide
+
+## Security vuln disclosure - perhaps in SECURITY.md?
```
koivunej: I think this could be just clear policy to use issues, which I guess is the default and the one for rust-libp2p which doesn't have a CONTRIBUTING.md.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
+# Contributing to Rust IPFS
+
+Welcome, and thank you for your interest in contributing to Rust IPFS. Issues and pull requests are encouraged.
+
+## Target Build
+
+Rust IPFS will always target the current _stable_ version of Rust that is released. Our CI/CD tests will reflect this.
+
+TODO: Instructions on installing the up-to-date stable toolchain
```
koivunej: The book discusses this at length, I think in general it'd best to always refer it: https://doc.rust-lang.org/book/ch01-01-installation.html. Now that I finally read this chapther I always link, I wished it discussed other options as well, but the current form is good enough.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Link to roadmap issue or project?
+
+## Contributing
+
+Reporting Bugs
+Suggesting Enhancements
+Your First Code Contribution
+Pull Requests
+
```
koivunej: I'd like to add here "documentation fixes and improvements" or similar. I think "open for doc prs" is quite important as a sign of appreciating such. I think that while there is little documentation at the moment, welcoming "usage issues" as could be a good idea. Though, in general I go to this file when wanting precise information on the tactics of the project, how exactly should I fill the bug report in. Perhaps my ideas are more geared towards README?

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+## TODO: Styleguides
+
+Git Commit Messages
+JavaScript Styleguide
+CoffeeScript Styleguide
+Specs Styleguide
+Documentation Styleguide
+
```
koivunej: There isn't uniform commit message style, like, I try to use the "[angularjs style](https://gist.github.com/stephenparish/9941e89d80e2bc58a153)" but make mistakes and the same time wouldn't want to start checking the messages like `js-ipfsd-ctl` does for example.

I think this section could be replaced with discussion on what the CI pipeline does: It runs `cargo fmt` and `cargo clippy`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+## Target Build
+
+Rust IPFS will always target the current _stable_ version of Rust that is released. Our CI/CD tests will reflect this.
+
+TODO: Instructions on installing the up-to-date stable toolchain
+TODO: First Principles
+
+## Roadmap
+
+- Link to roadmap issue or project?
```
koivunej: Here I am thinking the strongest on "this should be in readme". If not, obvious one is the grant to link here (with the permalink, because there might be renames/reorgs and I'd hate for someone having to hunt down broken links): https://github.com/ipfs/devgrants/blob/52232fd6516a6d08db698eee62ea86f7dba5f7ad/open-grants/ipfs-rust/README.md

CONTRIBUTING.md
vmx: From the AngularJS style message comes the https://www.conventionalcommits.org/. The nice thing about it is that you can generate changelogs from them. That's really nice especially for breaking changes.

CONTRIBUTING.md
```diff
@@ -0,0 +1,37 @@
...
+
+- Reporting Bugs
+- Suggesting Enhancements
+- Adding new tests
+- Adding new functionality
+- Documentation-only udpates
+
+## Style
+
+- Git Commit Messages should lean towards Conventional Commits but will not be enforced
```
koivunej:
```suggestion
- Git Commit Messages should lean towards [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) but will not be enforced
```

CONTRIBUTING.md
koivunej: I didn't know a spec had been formed, probably worth linking to it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Contributing
+
+We welcome all forms of contribution. Please open issues and PRs for:
+
+- Reporting Bugs
+- Suggesting Enhancements
+- Adding new tests
+- Adding new functionality
+- Documentation-only udpates
```
koivunej:
```suggestion
- Documentation-only updates
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,37 @@
...
+* Implementing endpoints and features not covered in the dev grant (See the [README](./README.md#roadmap))
+* Tests and CI for existing functionality
+* Examples
+* Documentation
+
+## First Principles
+
+1. Keep the build time small
+2. Aim for high (but not absolute) code coverage in testing
+3. Keep the dependency count low (use [`cargo-deps`](https://github.com/m-cat/cargo-deps))
```
koivunej: I am thinking there is likely a lot of work to do ahead and adding these principles to minimize the build time might complicate it. I feel that these are good principles except for the dependency count, which could transitively give you better build times but that might be better as a later optimization...? Building a dependencyless ipfs node is possible but community efforts should be leveraged, so I feel there is a conflict.

Otherwise these wouldn't matter, but I feel like if I approve these, I should make sure my next PR would probably at least more difficult to do. Not sure what to write here, or if anything should. Now that something has been offered, I feel like we should have something like this. Any other ideas?

aphelionz: I'll remove the dependency count line for now and other people can chime in. We'll learn more as we go and actively collaborate

aphelionz:
@dvc94ch OK to merge from your perspective?

dvc94ch:
Nice work!

aphelionz:
Why thank you!

PR #69, opened 2020-03-04
CI running on push to master and PRs

aphelionz:
I noticed that CI was running twice if I created a branch and made a PR. How do we feel about this new config?


vmx:
Having CI run on push is nice if you just want to have tests running via the CI, but you are not ready for doing a PR yet. Though I don't think it's super important.

PR #71, opened 2020-03-05
Http v0 for testing

### LINKED ISSUE #60: HTTP v0 API for testing
Per the grant plan testing with [`interface-js-ipfs-core`](https://github.com/ipfs/interface-js-ipfs-core/) and [`interop`](https://github.com/ipfs/interop/) requires an [HTTP v0 API](https://docs-beta.ipfs.io/reference/http/api/). The launch an ipfs "daemon" part of the two test resources is in [`js-ipfsd-ctl`](https://github.com/ipfs/js-ipfsd-ctl).

I understand the "daemon" to be just a process like go-ipfs `daemon` which will hang in the foreground until `POST /shutdown` has been initiated, but this should be verified with the tests of `js-ipfsd-ctl`.

The API endpoints we need in phase 1.0 can all return "501 Not Implemented" except for `/shutdown` which should allow us to stop the process gracefully. The test cases of js-ipfsd-ctl expect `/id` to return something.

Command line option `init` should do similar directory setup and keygen as `go-ipfs init` at the `IPFS_PATH` set by the environment. During normal startup there can be other arguments but after listening has been started, api listening address needs to be print `API listening on: $multiaddr` and finally a `daemon is running` line must be output.

`init` should produce an empty file called "api" inside the `IPFS_PATH`, which should be written to when the listening has started with the `$multiaddr` and on exit the file should be truncated.

In channel discussions a `rust-ipfs/{http,cli}` was suggested and adding such to the root `Cargo.toml` workspace and a relative dependency seems to work.

koivunej:
Latest progress with #60: the http binary and sub-crate simply trying to find a way to be compatible with [`js-ipfsd-ctl`](https://github.com/ipfs-rust/js-ipfsd-ctl/) and then with [`interface-js-ipfs-core`](https://github.com/ipfs-rust/interface-js-ipfs-core) in [`ipfs-rust-conformance`](https://github.com/ipfs-rust/ipfs-rust-conformance) and finally with [`interop`](https://github.com/ipfs-rust/interop).

Bad thing about the subcrate is the two Cargo.lock files need to get updated at all times.

Open problems:

 * solving the fact that `interface-js-ipfs-core` and `interop` use different versions (1.0.0, 0.47.4) and I started working on 3.0.0.
 * `ipfs-rust-conformance` seems to work with the patched `js-ipfsd-ctl` 3.0.0 (we get proper test failures for the select tests)
 * how to get any and all of these executed in CI compatible time (`npm install` in `ipfs-rust-conformance` took 450s)

Closes #60.


vmx:
FYI (no need to change it): I've good experience with https://crates.io/crates/flexi_logger for logging. It's more flexible than env_logger, though still pretty simple. E.g. you can easily pass in your custom logging style if you e.g. need thread IDs for debugging purpose.

aphelionz:
I think we can likely merge this, since it stands independently of the node.js testing framework that would be running against it.

By two Cargo.lock files you mean the one at the root `/Cargo.lock` and the one in the sub-crate `/http/Cargo.lock`?

koivunej:
> I've good experience with https://crates.io/crates/flexi_logger for logging.

The env_logger was already a dependency used in the `rust-ipfs` and didn't want to think too much here. Thanks for mentioning `flexi_logger`, hadn't heard about it and it looks quite feature-full. My  long-term idea for logging has been `tracing` with it's future support and "mdc" features, but I guess we'll need juicier bugs before that is warranted :)

koivunej:
I guess we can just make this non-draft, hopefully attracting some reviews, not that there is much interesting yet. I can't see a CI solution yet but perhaps this should be postponed until the thing does something useful.

> By two Cargo.lock files you mean the one at the root /Cargo.lock and the one in the sub-crate /http/Cargo.lock?

Yes.

aphelionz:
Yes, merging OK with me. At worst I can manually run `ipfs-rust-conformance` and create the reports by hand.

vmx:
> Bad thing about the subcrate is the two Cargo.lock files need to get updated at all times.

Why does the subcrate need a lock file? Usually you have only a single lock file at the root of the workspace.

.gitignore
```diff
@@ -1,2 +1,2 @@
-/target
+**/target
```
dvc94ch: this shouldn't be needed, since you added the crate to the workspace it'll use the workspace target directory and you can remove the crate target directory on you fs.

dvc94ch: please also remove the redundant Cargo.lock

http/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "http"
+version = "0.1.0"
+authors = ["Joonas Koivunen <joonas@equilibrium.co>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
dvc94ch: comment is unnecessary

http/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
```
dvc94ch: I like ordering deps alphabetically as it helps readability once they're 10-20 dependencies

aphelionz:
Thanks @dvc94ch I'll get on these updates right now

.gitignore
aphelionz: Should be all set, reverted the `.gitignore` and removed the Cargo.lock

http/Cargo.toml
aphelionz: Removed

http/Cargo.toml
aphelionz: Reordered

koivunej:
Thanks for @aphelionz for the cleanup and @dvc94ch and @vmx  for spotting the issues. I think I messed this up by adding the subcrate, going ahead and working on it before adding it to the workspace. I must've observed `Cargo.lock` getting modified after initial adding so I wrongly deduced it should be kept.

http/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "http"
```
aphelionz: We probably need to name this something less generic like `rust-ipfs-http` or some such thing because of https://crates.io/crates/http

koivunej:
```suggestion
name = "rust-ipfs-http"
```

This would be in the similar fashion as other subcrates I've seen, like for example rust-libp2p.

PR #73, opened 2020-03-05
Test README on ci.

### LINKED ISSUE #41: Try skeptic for keeping the README up to date
[`skeptic`](https://github.com/budziq/rust-skeptic) can compile the code inside markdown files (README.md). There was an issue when nightly rustc was used but now it might work with stable rustc.

Tagging as good first issue as this shouldn't be painful this time. If you are interested and run into issues or need help please ask on this issue.

dvc94ch:
Ended up not using skeptic as it didn't work properly.

Fixes #41


aphelionz:
I see you switched to `nightly` on CI. Does nightly have skeptic built in or some such thing? Will that still allow us to honor our goal of supporting `stable`? I think you can use the `matrix` setting in GH Actions to test both, but if skeptic being built in is the case then that might be problematic.

koivunej:
> Does nightly have skeptic built in or some such thing?

This is with [`#![feature(external_doc)]`](https://doc.rust-lang.org/unstable-book/language-features/external-doc.html) which I did not know.

> Ended up not using skeptic as it didn't work properly.

I hadn't tried it after we started using stable, was it a linker error again? 

I feel like keeping on stable is perhaps even more important than having a readme example, the examples/readme.rs could also be linked?

aphelionz:
What exactly didn't work with https://github.com/budziq/rust-skeptic?

dvc94ch:
Didn't find the dependencies. The current solution doesn't prevent using it on stable, as the example is only tested when the nightly feature is enabled.

aphelionz:
Gotcha, that's too bad that it couldn't find the dependencies. What about creating a second file, sibling to ci.yml that contains this particular workflow and then keeping the mainline "Rust IPFS" action as-is. The former nightly, the latter stable. That way we cover all the cases and still get README testing support until that becomes stable as well?

aphelionz:
LGTM :+1:

PR #98, opened 2020-03-12
Updating ci.yml to auto-run conformance testing suite

aphelionz:


.github/workflows/ci.yml
```diff
@@ -155,3 +155,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Cache cargo folder
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo
+          key: readme-doctest
```
dvc94ch: should have a different key, the readme-doctest are nightly artifacts which won't do anything if you're working with stable

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          rust-version: stable
+
+      - name: Build
+        if: matrix.platform.cross == false
+        run: cargo build --workspace
+
+      - name: npm install
+        run: npm install
+        working-directory: ./conformance
```
koivunej, line 85: Results of the `npm install` should likely be cached right away.

We likely don't want this on platforms which cannot build `ipfs-http` or platforms which can build it (possibly android) but we'd require an interpreter to run the tests in (cross compiled).

As the same goes for the next three tasks, I was thinking that maybe these tasks should be limited to only certain platforms (rust tier 1 for example). Another job would be great, if it re-use the `ipfs-http` from the previous job but I don't know how to do that.

PR #100, opened 2020-03-13
Switching to 64 bit iOS target

aphelionz:


koivunej:
I see this must've been failing the latest commits, merging in.

PR #104, opened 2020-03-17
Implement swarm api.

dvc94ch:


aphelionz:
Looks pretty good at first glance but noticed that no new tests were written. Will those come with the final PR?

dvc94ch:
I only opened the pr because @koivunej asked me to, just got my battery replaced, so should get some work done tomorrow

PR #111, opened 2020-03-22
Refactor options

dvc94ch:
I'll fix the tests and examples if @koivunej approves

There's a bunch of other needed refactors:
- Introduce an ipfs daemon similar to what I proposed in a previous PR
- Make the strategy a network behaviour and decouple it from bitswap
- Turn the repo into a network behaviour
- Drop the dag/path and custom store traits in favor of libipld


src/options.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        write!(fmt, "Keypair::{}", kind)
+    }
+}
+
+impl<I: Borrow<Keypair>> AsRef<Keypair> for DebuggableKeypair<I> {
+    fn as_ref(&self) -> &Keypair {
+        self.0.borrow()
+    }
+}
```
koivunej: I am glad you found a way for this to work, I feel I must've tried out this configuration as well but always ran into "needs type hint" issue!

src/lib.rs
```diff
@@ -192,7 +65,7 @@ type Channel<T> = OneshotSender<Result<T, Error>>;
 /// Events used internally to communicate with the swarm, which is executed in the the background
 /// task.
 #[derive(Debug)]
-enum IpfsEvent {
+pub enum IpfsEvent {
```
koivunej: Why does this need to be public? Maybe `pub(crate)`?

src/options.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// The path of the ipfs repo.
+    pub ipfs_path: PathBuf,
+    /// The keypair used with libp2p.
+    pub keypair: Keypair,
+    /// Nodes dialed during startup.
+    pub bootstrap: Vec<(Multiaddr, PeerId)>,
+    /// Enables mdns for peer discovery when true.
+    pub mdns: bool,
+}
+
```
koivunej: The `fn inmemory_generated_keys(mdns: bool) -> IpfsOptions<TestTypes>` is missing. I am fine with other names but there needs to be a way to create an IpfsOptions for tests easily, without any IO. I'm not sure if the `Default` impl should do that, maybe not change the semantics at least yet?

src/lib.rs
dvc94ch: Doesn't compile otherwise, because the repo takes a `Sender<IpfsEvent>`, it's leaking private types.

dvc94ch: I'll fix this in a follow up when turning the repo in to a network behaviour. Then we should also be able to remove all the `Arc<Repo>`.

koivunej: Aah, so the real cause is Repo having `pub fn new(...) -> Self`? That'd would allow `pub(crate)` here as well.

dvc94ch:
That is left for future PR's, not part of this one.

dvc94ch:
@koivunej it would require making the new method on repo pub(crate) too. I'd prefer to move the rest of lib.rs to daemon.rs, make it all pub and then only export the stuff we want to export in lib.rs, does that sound ok to you?

dvc94ch:
Superceeded by #119 which includes all of the above mentioned

Issue #11, opened 2019-02-27
Pinning support


dvc94ch:
Related issue is garbage collection of unpinned blocks

saresend:
Would there be guidance available for this? I'd be interested in tackling it

dvc94ch:
I guess the best is to dive in. What sort of guidance did you have in mind?

You probably want to have a table in the datastore `Cid -> bool`. There's a utility in libipld to find all cids recursively, since supporting recursive pinning is very useful. We might drop rocksdb in the future in favor of sled.

If you have an ipld datastructure you will want to pin the root. After you make a change you want to pin the new root and unpin the old one. We might end up requiring two different types of pinning, as when streaming for example we only the buffered part of the file to be pinned. There we want to manage individual blocks manually.

koivunej:
Related: the discussion in #84 and the ipfs/specs discussion linked from there.

saresend:
I didn't have anything too specific in mind, just wanted to check that you guys would be willing to support the changes since I'm still ramping up on IPFS stuff! I'll take a look and hopefully have a PR up soon

aphelionz:
Great @saresend and welcome! We look forward to it.

saresend:
Have something very tentative up at #117, would be interested in seeing if the direction here is correct!

aphelionz:
@dvc94ch Should we leave this open to capture the remaining work or close and create new issue(s)?

koivunej:
For posterity: The testing of #90 requires `/pin/add` so it will likely be added during that, for the conformance tests. If I'll implement it I'll keep it as constrained as possible (for example, non-recursive if allowed by the tests, which I understand only require single block pinning).

saresend:
Hey, sorry that I haven't got around to it just yet, expect a PR up in the next couple of days! @koivunej Let me know if you want to take that work, don't want to duplicate if you're already working on it.

koivunej:
@saresend I'll be soon (hopefully!) moving over to this. I want to handle #238 first though and lets just say it hasn't been so fast moving as I would had hoped. After #238 moving `refs` over to `ipfs` for different kinds of pins should start to move forward.

Regarding #152. If you want, we could iron out what exists at the moment and get it merged in. Let me know.

saresend:
Ah, so sorry just saw this. I've started work and been very busy. I should have some time later this week however, if there hasn't been too much development on this front since!

aphelionz:
[...]

PR #117, opened 2020-03-23
Adds basic support for pin and unpin operations

saresend:
## What
Should address #11, by adding some basic pin / unpin operations to the dag

## Todo
- [x] Add support for querying whether a block is pinned
- [ ] Add Tests once design is finished


aphelionz:
This is really exciting, thank you @saresend. If this works for @koivunej then it'd be great to see tests, and then if you want to go for the gold, add the http endpoints! 🎉

dvc94ch:
I guess remove should check if the block is pinned.

saresend:
Alright, a couple of changes: 
* First, pinning is defined by the existence of a pair in the Pin hashmap, rather than a boolean flag of 1 or 0. This way we don't have two `False` states, in the form of `Null` or `Zero`, and instead consolidate all into the null case.

* It should now also error when trying to unpin an unpinned block, rather than just silently performing a no-op.

aphelionz:
@saresend Here's the overall spec, including what the conformance tests expect for the PIN apis. I hope this is helpful. Let me know if you have any questions

https://github.com/ipfs-rust/ipfs-rust-conformance/issues/2

saresend:
I was going to add some tests at the very least. I might leave the actual `http` endpoints to a separate PR however. If you have some code points for how you write tests for this project, I'd be happy to add them.

aphelionz:
@saresend @dvc94ch Tests and doc would be important. You can add simple module tests at the end of the respective files, e.g. `dag.rs` and `repo/mod.rs`. It would also be huge if you could do the endpoints. There's more examples of testing in that crate as well.

Thank you so much again for this.

saresend:
No problem, I'll open up a new PR for the http endpoints, just to keep concerns separate!

dvc94ch:
It's also not clear why you're exposing it on the dag, why unpin should error if a block already is unpinned, as stuff may share blocks or how this works as you can still remove a block even if it's pinned.

saresend:
Totally fair, I'm still ramping up on IPFS and don't have great context on how you envision this being architected or how concerns are being separated. If you have suggestions on better ways to expose the API I'm all ears. My understanding of pinning is that it protects against garbage collection, and I wasn't able to find whether garbage collection is already being done in `rust-ipfs`. My thought was that garbage collection could leverage the pinning query logic when removing nodes whenever it does get built

aphelionz:
As far as the error goes, it should simply follow what js-ipfs or go-ipfs does. So if they error when trying to unpin and already unpinned block, then we should too. Otherwise not.

dvc94ch:
I don't think the dag should expose pin/unpin, and if it did then the semantics would probably not be to pin the root of a path as a path can point to other blocks, so you'd want to pin the last block it refers to.

Yes pin is for garbage collection, which is not currently implemented. What currently is implemented is manually removing blocks, which shouldn't happen if it's pinned.

Since blocks can be shared, pinning should probably increment a counter and unpin should decrement it, instead of having a boolean.

Garbage collection can be implemented by iterating through all blocks, checking if the pin count is zero and if it is remove the block.

saresend:
Where would you recommend pinning be implemented?

src/lib.rs
```diff
@@ -312,6 +312,21 @@ impl<Types: IpfsTypes> Ipfs<Types> {
         Ok(self.repo.remove_block(cid).await?)
     }
 
+    /// Pins a given Cid
+    pub async fn pin_block(&self, path: IpfsPath) -> Result<(), Error> {
```
dvc94ch:
```suggestion
    pub async fn pin_block(&self, cid: &Cid) -> Result<(), Error> {
        self.repo.pin_block(cid).await
    }
```

aphelionz:
I might suggest we focus on tests and the HTTP endpoints here, so that at least there's a foundation for architecture changes + refactors. Always room for improvement, especially if we keep getting awesome community contributions.

dvc94ch:
Do you have anything to add in regards to my comment? Did I not explain in which ways the implementation suggested here is flawed clearly enough? Conformance tests are not a way to avoid using your brain, as tests are just as buggy as code. Besides tests do not ensure correctness, they sometimes show the presence of bugs.

dvc94ch:
Let me try again, I'll use letters to denote blocks in the following diagram:

```txt
a -> b <- c
```

and these two concurrent processes running in parallel

```txt
process 1:
    put(b)
    pin(b)
    put(a)
    pin(a)
    ...
    unpin(a)
    unpin(b)
    gc()
```
```txt
process 2:
    put(b)
    pin(b)
    put(c)
    pin(c)

    get(b)???
```

dvc94ch:
Also put should probably pin automatically to prevent a race, but that's something to be discussed, if someone can supply a usecase where you don't want to put_and_pin

dvc94ch:
I approved the PR and would have merged it if it didn't have conflicts with master. But I agree with @aphelionz, this features flaws should be extensively documented and tested.

saresend:
Before the next couple changes, I just want to clarify what you'd want for this PR. So far, the changes requested are:
* Move to use `&Cid`, rather than `ipfsPath`
* Automatically Pin on Put actions
* Prevent removal based on whether a block is pinned
* Add tests and documentation

Is there anything I'm missing here?

dvc94ch:
If we didn't have the conformance tests I'd suggest removing remove altogether. Also don't use a bool as shown in the example above it can lead to issues. But mainly I'm a bit annoyed that my approval on pr's has been purchased for a mere 2000€, which has nothing to do with your pr. Thanks for contributing...

saresend:
@dvc94ch @aphelionz, alright this should include a more `RC` like logic for pinning, removes it from the Dag API, and finally includes a test to verify behavior. It should also now be rebased so if it all checks out feel free to merge.

dvc94ch:
Looks good! So I think this is much better, thanks. Can you rebase again, looks like it's not mergeable.

saresend:
Alright that should be merged!

koivunej:
Interesting the same test is flaky on windows ... Again. Need to investigate why.

koivunej:
The test failure will be fixed with #124 but I think this is now good to merge?

aphelionz:
@saresend Please add yourself to https://github.com/ipfs-rust/welcome/issues/1 and also let us know if you plan on doing the endpoints as well

saresend:
Yep! I plan to have a PR up in the next week or so, classes have started back up again for me so it may take me a bit longer unfortunately

aphelionz:
That's ok, no rush.

PR #126, opened 2020-03-27
Bitswap endpoint

dvc94ch:


dvc94ch:
it's not about limiting concurrency. it's that it's not really concurrent in the first place, we don't spawn any tasks other than in the strategy (and the tasks that libp2p spawns internally), and the tasks spawned in the strategy are simply overengineering as currently the strategy doesn't do anything interesting. any long running tasks should be spawned (the block store actually is spawning tasks now making it concurrent). it's essentially the same design as a warp or any other http server. their design is for every request a task is spawned. the ipfs handle is the client to the server, which spawns tasks when appropriate.

dvc94ch:
Essentially what this pr tries to accomplish was the following:

- Introduce an ipfs daemon similar to what I proposed in a previous PR
- Make the strategy a network behaviour and decouple it from bitswap
- Turn the repo into a network behaviour
- Drop the dag/path and custom store traits in favor of libipld

with the better understanding I gained from trying to implement this these things slightly changed, but the ideas are the same

dvc94ch:
the way I was planning to increase concurrency in the ipld-deamon is by using the fact that file systems are already concurrent. a client could read a block from the file system and cache it in memory directly. a request to to the ipld-deamon would only be needed to write a block, with the corresponding buffer to minimize writes to disk by only writing to disk on flush or when a block is not found locally request it to be found with bitswap. that design however was decided not to be pursued at the beginning of the grant.

PR #128, opened 2020-03-28
set openssl dir and enable cross-compile

### LINKED ISSUE #96: Partial builds on CI cross targets
During #95 I excluded building `rust-http-ipfs` on cross targets (incl. android). Fixing this probably isn't anything else than to find/add the prebuilt `openssl-sys` friendly packages for the cross targets. Could also be that the env variable just needed to be set, as mentioned by `openssl-sys` build script.

See https://github.com/ipfs-rust/rust-ipfs/runs/503319261 where the android builds failed and macos cross builds were not actually run because I fumbled `--target xyz` away from the build command by mistake.

The logs are not loading as of writing this so hopefully they'll work later on.

ifkb99:
fix #96 along with including ipfs-http

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
       LLVM_CONFIG_PATH: /usr/local/opt/llvm/bin/llvm-config
       NDK_HOME: /usr/local/lib/android/sdk/ndk-bundle
       VCPKGRS_DYNAMIC: 1
+      PKG_CONFIG_ALLOW_CROSS: 1
 
     runs-on: ${{ matrix.platform.host }}
     steps:
@@ -105,8 +106,10 @@ jobs:
 
     - name: Build other cross compilations
       if: contains(matrix.platform.target, 'android') == false && matrix.platform.cross == true
-      run: cargo build --workspace --exclude ipfs-http --target ${{ matrix.platform.target }}
-      # exclude http on other cross compilation targets because openssl
+      run: cargo build --workspace --target ${{ matrix.platform.target }}
+      env:
+        OPENSSL_LIB_DIR: '/usr/lib/x86-64-linux-gnu/libssl.a'
+        OPENSSL_INCLUDE_DIR: '/usr/include/openssl'
 
     - name: Rust tests
       if: matrix.platform.cross == false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,13 +36,13 @@ jobs:
           host: ubuntu-latest
           cross: true
 
-        - target: x86_64-apple-ios
-          host: macos-latest
-          cross: true
+      # - target: x86_64-apple-ios
+      #   host: macos-latest
+      #   cross: true
 
-        - target: aarch64-apple-ios
-          host: macos-latest
-          cross: true
+      # - target: aarch64-apple-ios
+      #   host: macos-latest
+      #   cross: true
 
        # - target: wasm32-unknown-unknown
        #   host: ubuntu-latest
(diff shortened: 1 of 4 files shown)


ifkb99:
@aphelionz Should I keep trying directories that openssl may be in, or is there a better method?
https://github.com/ipfs-rust/rust-ipfs/pull/128/checks?check_run_id=541078376#step:11:205

aphelionz:
There's almost certainly be a better method, but I'm not yet familiar enough with the GH Actions runners to say - let me take a look at a few things. Also @hugomrdias posted some helpful links in Matrix: https://matrix.to/#/!JxzWpvsvfYuMHyUJiL:matrix.org/$t3B9RULiM-kSkPDR31xVtCiR4sAbeClHpteHijFgZyg?via=matrix.org&via=hielle.com&via=t2bot.io

aphelionz:
@ifkb99 Can you try: 

1. Adding this crate to `http/Cargo.toml`: https://crates.io/crates/openssl-probe
2. Adding the line from `openssl-probe`'s README to the `main` function in `http/src/main.rs`

aphelionz:
You'll also want to merge master in because of #124

aphelionz:
Whoa, so close! Now what to do with the ios builds :thinking:

koivunej:
The windows test failure is one of the tests still racy to access the configuration file, the doctest failure is the flaky test fixed in #124 which would go away on rebasing to master.

Looks like you got the android build at least to work! Could leave the issue open for the mac ones since this is progress already! Though the android build apparently doesn't run tests, since tests currently [guarded with `cross == false`](https://github.com/ipfs-rust/rust-ipfs/blob/master/.github/workflows/ci.yml#L111-L113). I don't see why tests could be be run on android, but now the separation from "cross[compiled]" will need to change to "apple" or something else.

koivunej:
Rerunning the jobs just in case, looks like you merged back to your branch in the latest commit.

aphelionz:
I'd be ok with disabling the ios builds as we have before this, and then opening an issue to address them separately 👍

koivunej:
Yeah, to recap:

 * exclude http on ios

I suggested running the tests on andoid but that of course wont work without a virtualization environment which I don't think we want to add here.

aphelionz:
Yes, I was agreeing

aphelionz:
@ifkb99 Can you disable the builds on the ios versions so we can get a green CI again? Then we can merge. Also, please comment here to add yourself to the roster: https://github.com/ipfs-rust/welcome/issues/1

aphelionz:
Looks like this needs a rebase and you should be good to go

.github/workflows/ci.yml
```diff
@@ -36,13 +36,13 @@ jobs:
...
+      # - target: x86_64-apple-ios
+      #   host: macos-latest
+      #   cross: true
 
-        - target: aarch64-apple-ios
-          host: macos-latest
-          cross: true
+      # - target: aarch64-apple-ios
+      #   host: macos-latest
+      #   cross: true
```
koivunej, line 45: Doing this will disable the builds altogether while we could just exclude the problematic `ipfs-http` on *these platforms*. Will comment suggestion below with the relevant build step.

.github/workflows/ci.yml
```diff
@@ -105,8 +106,7 @@ jobs:
 
     - name: Build other cross compilations
       if: contains(matrix.platform.target, 'android') == false && matrix.platform.cross == true
-      run: cargo build --workspace --exclude ipfs-http --target ${{ matrix.platform.target }}
-      # exclude http on other cross compilation targets because openssl
+      run: cargo build --workspace --target ${{ matrix.platform.target }}
```
koivunej, line 109: [Instead of commenting out the IOS builds](https://github.com/ipfs-rust/rust-ipfs/pull/128/files#r404651653) I would prefer if this line would just restored to what it was earlier:

```
run: cargo build --workspace --exclude ipfs-http --target ${{ matrix.platform.target }}
```

This would mean that IOS builds are still without `ipfs-http` if the openssl is still problematic after adding the `openssl_probe`. However if the `openssl_probe` now allows building  ipfs-http on android, the line 104 could have this `--exclude ipfs-http` removed.

koivunej:
Closing this for inactivity.

PR #137, opened 2020-04-05
Implements dag-put and block-stat

dvc94ch:


.cargo/config
```diff
@@ -0,0 +1,3 @@
+[build]
+#rustflags = ["-C", "linker=clang"]
+rustflags = ["-C", "link-arg=-fuse-ld=lld"]
```
koivunej, line 3: I don't think we should do this in the project root but hint at this in the README (using `~/.cargo/config` and same contents) as the lld doesn't ship with rustc at least yet.

http/Cargo.toml
```diff
@@ -6,26 +6,25 @@ edition = "2018"
...
 
 [build-dependencies]
-vergen = "3.0.4"
 prost-build = "0.6.1"
+vergen = "3.1.0"
 
 [dependencies]
-log = "0.4"
-env_logger = "*"
+base64 = "0.12.0"
```
koivunej, line 13: Do we need both base64 and multibase (it has `multibase::Base::Base64Pad` for what base64 generally means). I used the latter in pubsub (all fields needed to be base64).

dvc94ch: It's not used at all, since the api expects base58 but looks like dag-json which mandates base64

http/src/v0/dag.rs
```diff
@@ -0,0 +1,74 @@
...
+        .data()
+        .await
+        .ok_or(InvalidMultipartFormData)?
+        .map_err(|_| InvalidMultipartFormData)?;
+    let data = buf.to_bytes().as_ref().to_vec().into_boxed_slice();
+    let digest = hasher(&data);
+    let cid = Cid::new_v1(format, digest);
+    let reply = json!({
+        "Cid": { "/": cid.to_string() }
+    });
```
koivunej, line 60: good idea!

aphelionz:
@dvc94ch What is needed to get CI back to green?

Cargo.toml
```diff
@@ -11,25 +11,25 @@ nightly = []
...
 async-std = { version = "1.5.0", features = ["attributes", "std"] }
-async-trait = "0.1.24"
+async-trait = "0.1.29"
 bitswap = { path = "bitswap" }
 byteorder = "1.3.4"
 dirs = "2.0.2"
 domain = { git = "https://github.com/nlnetlabs/domain", features = ["resolv"] }
 futures = { version = "0.3.4", features = ["compat", "io-compat"] }
-libipld = { version = "0.1.0", features = ["dag-pb"] }
+libipld = { version = "0.1.0", features = ["dag-json", "dag-pb"], path = "../rust-ipld" }
```
koivunej, line 22:
```suggestion
libipld = { git = "https://github.com/ipfs-rust/rust-ipld", features = ["dag-pb"] }
```
Took me a while to understand the CI failure, but I happened to have a recent checkout in the same path :)

aphelionz: Aha!

aphelionz: Forgive me @dvc94ch but I am going to commit this so I can run conformance testing on this branch

vmx: @aphelionz there's also a way to patch a dependency https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section which might be a clearer indicator of doing something temporarily.

aphelionz:
@dvc94ch checking in here again, what's the plan for getting this merged? It looks like the feedback is minimal and the CI just needs to be fixed

dvc94ch:
I think @koivunej took over

.cargo/config
aphelionz: I had suggested @dvc94ch do this, my bad. I'll remove for now and add it to the README.

aphelionz:
Everything looks good here except for one test in `dag.put`

> 18) should override hash algorithm default and resolve with it

PR #150, opened 2020-04-10
Refs local endpoint

aphelionz:
This PR adds a conformance-passing `/api/v0/refs/local` endpoint.

I added a `list` function to the `BlockStore` trait.

TODO:
- [x] There's probably too much `Result`, but maybe not. It seems like there are possibilities for failure at many steps
- [x] Speaking of that, error handling is needed as well. Looking at how to deal with that and clean things up.

Note that this includes the work from #137 and will be rebased

### DIFF
--- a/http/src/v0/refs.rs
+++ b/http/src/v0/refs.rs
@@ -1,0 +1,92 @@
+use futures::stream;
+use ipfs::error::Error;
+use ipfs::{Ipfs, IpfsTypes};
+use serde::Serialize;
+use warp::hyper::Body;
+use warp::{path, Filter, Rejection, Reply};
+
+use super::support::{with_ipfs, StringError};
+
+#[derive(Serialize, Debug)]
+struct RefsResponseItem {
+    #[serde(rename = "Err")]
+    err: String,
+
+    #[serde(rename = "Ref")]
+    refs: String,
+}
+
+/// Handling of https://docs-beta.ipfs.io/reference/http/api/#api-v0-refs-local
+pub fn local<T: IpfsTypes>(
+    ipfs: &Ipfs<T>,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path!("refs" / "local")
+        .and(with_ipfs(ipfs))
+        .and_then(inner_local)
+}
+
+async fn inner_local<T: IpfsTypes>(ipfs: Ipfs<T>) -> Result<impl Reply, Rejection> {
+    let refs: Vec<Result<String, Error>> = ipfs
+        .refs_local()
+        .await
+        .map_err(StringError::from)?
+        .into_iter()
+        .map(|cid| cid.to_string())
+        .map(|refs| RefsResponseItem {
+            refs,
+            err: "".to_string(),
+        })
+        .map(|response| {
+            serde_json::to_string(&response).map_err(|e| {
+                eprintln!("error from serde_json: {}", e);
+                HandledErr
+            }).unwrap()
+        })
+        .map(|ref_json| Ok(format!("{}{}", ref_json, "\n")))
+        .collect();
+
+    let stream = stream::iter(refs);
+    Ok(warp::reply::Response::new(Body::wrap_stream(stream)))
+}
+
+#[derive(Debug)]
+struct HandledErr;
+
+impl std::error::Error for HandledErr {}
+
+use std::fmt;
+
+impl fmt::Display for HandledErr {
+    fn fmt(&self, _fmt: &mut fmt::Formatter) -> fmt::Result {
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::inner_local;
+    use ipfs::Block;
+    use libipld::cid::Cid;
+    use libipld::cid::Codec;
+    use multihash::Sha2_256;
+
+    #[tokio::test]
+    async fn test_inner_local() {
+        use ipfs::{IpfsOptions, UninitializedIpfs};
+
+        let options = IpfsOptions::inmemory_with_generated_keys(false);
+
+        let (mut ipfs, fut) = UninitializedIpfs::new(options).await.start().await.unwrap();
+        drop(fut);
+
+        for data in &[b"1", b"2", b"3"] {
+            let data_slice = data.to_vec().into_boxed_slice();
+            let cid = Cid::new_v1(Codec::Raw, Sha2_256::digest(&data_slice));
+            let block = Block::new(data_slice, cid);
+            ipfs.put_block(block.clone()).await.unwrap();
+        }
+
+        let _result = inner_local(ipfs).await;
+        // println!("{:?}", result.unwrap());
+    }
+}
--- a/src/repo/fs.rs
+++ b/src/repo/fs.rs
@@ -115,6 +115,12 @@ impl BlockStore for FsBlockStore {
         }
         Ok(())
     }
+
+    async fn list(&self) -> Result<Vec<Cid>, Error> {
+        // unwrapping as we want to panic on poisoned lock
+        let guard = self.cids.lock().unwrap();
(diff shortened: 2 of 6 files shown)


src/repo/fs.rs
```diff
@@ -115,6 +115,19 @@ impl BlockStore for FsBlockStore {
...
+
+    async fn list(&self) -> Result<Vec<Result<Cid, Error>>, Error> {
+        let cids: Vec<Result<Cid, Error>> = self
+            .cids
+            .lock()
+            .unwrap()
+            .to_owned()
+            .iter()
+            .map(|cid| Ok(cid.to_owned()))
+            .collect();
```
koivunej, line 127: Here I think is happening:

* `to_owned()` is actually HashSet::clone, which makes this more difficult to see
* `iter()` is called on the cloned temporary, after which
* `Ok(cid.to_owned())` will again clone the values

What would suffice is:

```rust
// unwrapping as we want to panic on poisoned lock
let guard = self.cids.lock().unwrap();
guard.iter()
    .map(|cid| cid.clone())
    .collect()
```

This would turn the signature from `Result<Vec<Result<Cid, Error>>, Error>` to `Result<Vec<Cid>, Error>`, also I am not sure what can fail here? I guess in reality we would be iterating the directories so I think it's good to keep the outer error, but I can't see any reason for the inner?

src/repo/mem.rs
```diff
@@ -61,6 +63,18 @@ impl BlockStore for MemBlockStore {
...
         Ok(())
     }
+
+    async fn list(&self) -> Result<Vec<Result<Cid, Error>>, Error> {
+        let blocks_mutex = self.blocks.lock().await;
+
+        let cids: Vec<Result<Cid, Error>> = blocks_mutex
+            .to_owned()
+            .iter()
+            .map(|(cid, _block)| Ok(cid.to_owned()))
```
koivunej, line 73: Same comments as before.

src/repo/mem.rs
```diff
@@ -61,6 +63,18 @@ impl BlockStore for MemBlockStore {
         self.blocks.lock().await.remove(cid);
         Ok(())
     }
+
+    async fn list(&self) -> Result<Vec<Result<Cid, Error>>, Error> {
+        let blocks_mutex = self.blocks.lock().await;
```
koivunej, line 68:
```suggestion
        let guard = self.blocks.lock().await;
```

This is the more common name. `blocks` would be fine as well, but calling it a mutex when it's not doesn't seem right.

src/repo/mod.rs
```diff
@@ -183,6 +184,10 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
         }
     }
 
+    pub async fn list_blocks(&self) -> Result<Vec<Result<Cid, Error>>, Error> {
+        Ok(self.block_store.list().await?)
+    }
```
koivunej, line 189: I was thinking that maybe this should be a passthrough method without the `async` but... It's difficult to say what is the lifetime of the returned value without `async fn` so I guess this is ok for now. Did you try this without `async fn`?

src/repo/mem.rs
```diff
@@ -164,6 +178,29 @@ mod tests {
...
+            let block = Block::new(data_slice, cid);
+            mem_store.put(block.clone()).await.unwrap();
+            assert!(mem_store.contains(block.cid()).await.unwrap());
+        }
+
+        let cids = mem_store.list().await.unwrap();
+        assert_eq!(cids.len(), 3);
+        for cid in cids.iter() {
+            assert!(mem_store.contains(cid.as_ref().unwrap()).await.unwrap());
+        }
```
koivunej, line 201: You could replace the `cids.iter()` with `cids` (take ownership of the values, as they are not needed after this loop) so that the next line `cid.as_ref().unwrap()` would become `&cid.unwrap()` and after the proposed removal of `Ok(cid...)` this would become `&cid`.

http/src/v0/refs.rs
```diff
@@ -0,0 +1,76 @@
...
+        .refs_local()
+        .await
+        .unwrap()
+        .iter()
+        // TODO: Error handling
+        // .map_err(|err| RefsResponseItem { refs: "", err: "" })
+        .map(|cid| cid.as_ref().unwrap().to_string())
+        .map(|refs| RefsResponseItem {
+            refs,
+            err: "".to_string(),
```
koivunej, line 39: the `"".to_string()` means "allocate a new string pls". I can fix this when reconciling my `refs` implementation, but while we are not really in a position to avoid all allocations, I think we should avoid the easy allocations.

This could not have the allocation if the field type was `Cow<'static, str>` which means "either a &'static str or a String", and here a `&'static str` of `""` would do just fine.

http/src/v0/refs.rs
```diff
@@ -0,0 +1,76 @@
...
+        })
+        .map(|response| serde_json::to_string(&response))
+        .map(|ref_string| Ok(format!("{}{}", ref_string.unwrap(), "\n")))
```
koivunej, line 42: It doesn't quite make sense for the json serialization to fail here. In case I had to debug such I'd prefer if the panic message would say something more relevant, like `ref_string.expect("Unexpected failure to serialize into json")`..

Note that `ref_string` is misleading variable name: the `ref_string` is not the `{ "Ref": "--> this value right here <--", ... }` but the whole json serialized, wrapped in a `Result<String, serde_json::Error>`, so maybe it could be `json` instead?

http/src/v0/refs.rs
```diff
@@ -0,0 +1,76 @@
...
+        .and(with_ipfs(ipfs))
+        .and_then(inner_local)
+}
+
+async fn inner_local<T: IpfsTypes>(ipfs: Ipfs<T>) -> Result<impl Reply, Rejection> {
+    let refs: Vec<Result<String, Error>> = ipfs
+        .refs_local()
+        .await
+        .unwrap()
+        .iter()
```
koivunej, line 33: The use of `into_iter()` would be preferred here as the Vec is owned.

http/src/v0/refs.rs
```diff
@@ -0,0 +1,76 @@
...
+        // TODO: Error handling
+        // .map_err(|err| RefsResponseItem { refs: "", err: "" })
+        .map(|cid| cid.as_ref().unwrap().to_string())
```
koivunej, line 36: This value is needlessly `Ok(cid)` wrapped at the blockstore implementations so removing that wrapping would alleviate this as well.

http/src/v0/refs.rs
```diff
@@ -0,0 +1,76 @@
...
+    path!("refs" / "local")
+        .and(with_ipfs(ipfs))
+        .and_then(inner_local)
+}
+
+async fn inner_local<T: IpfsTypes>(ipfs: Ipfs<T>) -> Result<impl Reply, Rejection> {
+    let refs: Vec<Result<String, Error>> = ipfs
+        .refs_local()
+        .await
+        .unwrap()
```
koivunej, line 32: This could be replaced with `.map_err(StringError::from)?` which would make a json response out of the error, and allow you to continue this chain with the value inside `Ok(_)`.

http/src/v0/refs.rs
```diff
@@ -0,0 +1,76 @@
...
+#[cfg(test)]
+mod tests {
+    use super::inner_local;
+    use ipfs::Block;
+    use libipld::cid::Cid;
+    use libipld::cid::Codec;
+    use multihash::Sha2_256;
+
+    #[tokio::test]
+    async fn test_inner_local() {
```
koivunej, line 58: This test looks incomplete...?

http/src/v0/refs.rs
koivunej: I'll handle this, resolving.

koivunej:
Test failures look transient...

aphelionz:
Fixed the commit msg :+1:

Issue #75, opened 2020-03-06
Use released dependencies


dvc94ch:
- [x] libp2p
- [x] cid
- [x] domain (waiting on decision async-std/tokio) #83 or remove while broken?
- [ ] ctr, possible workaround: #175 
- [x] ipld (not sure why we sticked with an unreleased but there have been heavy changes since)

aphelionz:
With regards to `domain`, it looks like in the graph in #63 that it uses tokio under the hood

dvc94ch:
We should probably stick to async-std, as it's what libp2p uses and try to find an async-std replacement for it.

aphelionz:
Seems so! Will need @koivunej to weigh in as the `http` subcrate has tokio in it at least once in its graph. Maybe we can start by keeping the core crate as close to async-std only as possible and work through the subcrates

koivunej:
> We should probably stick to async-std, as it's what libp2p uses and try to find an async-std replacement for it.

As rust-libp2p supports tokio as well, I don't see them using one. During the discussions we have so far I'm yet to see any other reason to use async-std, but offered a technical reason not to use it (the possibly upcoming scheduler). I would wish more technical reasons would be brought up instead of "X uses it" and I'd happy to learn from such.

While the split in the ecosystem is far from ideal, I think that crates around tokio are more mature for http for example. For DNS side of things in addition to `domain` there is [`trust-dns`](https://github.com/bluejekyll/trust-dns), while of course any resolution mechanism could be used via any threadpooling mechanism (as I understand async-std's `task::spawn_blocking` is at the moment).

Overall more of a subjective feeling I like the explicitness of tokio runtime starting as compared to implicit, which I feel is one of the reasons I like the rust language in general (being able to reason about code locally). As already mentioned in the chat, I think the ability to force using just a single threaded executor does help with implementing and debugging (I do not know if this can be forced with async-std) as the task starvation is more apparent. 

> [ ] ctr (by me)

The PR is still pending and I do not know how to make progress on that, other than querying progress every once in a while, I'll try to remember to do that. However if the difference was hit during testing, this would be a difficult one to debug, especially once we get interop tests running on CI.

Is there an overall pressing need to use released dependencies?

dvc94ch:
> the possibly upcoming scheduler

I'm not sure why that's a reason not to use async-std. Is there something wrong with that scheduler?

[...]

PR #151, opened 2020-04-10
fix: pin domain to specific revision

aphelionz:
This PR pins the domain crate to a specific revision. domain has moved on quite a bit and when this 
I tried to use `rust-ipfs` as a crate in another project, I got this error: 

```
error: failed to select a version for `domain`.
    ... required by package `ipfs v0.1.0 (https://github.com/ipfs-rust/rust-ipfs#89e3f7c5)`
    ...

the package `ipfs` depends on `domain`, with features: `resolv` but `domain` does not have these features.

failed to select a version for `domain` which could resolve this conflict
```

I'd love to get it updated to at least `0.5.0` but I don't know exactly how much work that entails. It does indeed raise the importance of #75

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -822,7 +822,7 @@ dependencies = [
 [[package]]
 name = "domain"
 version = "0.4.1"
-source = "git+https://github.com/nlnetlabs/domain#08496417cd2315970fdfb43201e619cac5e7a319"
+source = "git+https://github.com/nlnetlabs/domain?rev=084964#08496417cd2315970fdfb43201e619cac5e7a319"
 dependencies = [
  "domain-core",
  "domain-resolv",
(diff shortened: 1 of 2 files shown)


aphelionz:
@dvc94ch lmk if you can take a look. I think Joonas is on holiday. If you are as well, no worries

koivunej:
So why is this change necessary? The project builds just fine as it is for me, but is it because I already have a cached checkedout of their repository? Aah, I see, the issue happens when working with another lockfile. Makes sense. I'll rebase this and merge.

koivunej:
Why are there so many conflicts and changes in Cargo.lock...? When rebasing, I used the both Cargo.lock files, and did a `cargo check --workspace --tests` and I don't get as much changed. This is with 1.42.0.

koivunej:
let's merge as there's one green already

PR #153, opened 2020-04-14
Implementing /block/rm

aphelionz:
TODO:
- [x] Fix `test repo::mem::tests::test_mem_blockstore`
- [x] Fix `test repo::fs::tests::test_fs_blockstore`


src/repo/fs.rs
```diff
@@ -108,12 +109,13 @@ impl BlockStore for FsBlockStore {
 
     async fn remove(&self, cid: &Cid) -> Result<(), Error> {
         let path = block_path(self.path.clone(), cid);
-        let cid = cid.to_owned();
         let cids = self.cids.clone();
         if cids.lock().unwrap().remove(&cid) {
             fs::remove_file(path).await?;
+            Ok(())
+        } else {
+            Err(anyhow!("block not found"))
```
koivunej, line 117: Now looking this method as a diff I think that in general with blockstores we have (following the ideas from this great blog post: http://sled.rs/errors.html):

 - IO errors (filesystem broken, underlying media broken, underlying database is broken, background task inaccessible)
 - block removal result
   - not found
   - found but pinned
   - found and removed
   - pending download canceled

So I am thinking something like `Result<Result<(), BlockNotFound>, BadError>`, at least at this stage. At higher level this would probably need to transform to `Result<Result<BlockRemoved, BlockRemovalFailed>, BadError>` when we have the pins in play as well.

Perhaps:

```
enum BlockRemoved {
  Removed(Cid, usize),
  DownloadCancelled(Cid, Duration),
}

struct BlockNotFound(Cid);

enum BlockRemovalFailed {
  Pinned(Cid),
  NotFound(Cid),
}
```

Just thinking out loud for the post-anyhow times.

src/repo/mod.rs
```diff
@@ -199,8 +199,8 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
             .send(RepoEvent::UnprovideBlock(cid.to_owned()))
             .await
             .ok();
-        self.block_store.remove(cid).await?;
-        Ok(())
+
+        self.block_store.remove(cid).await
```
koivunej, line 203: I think we should add a TODO or FIXME to have `self.subscriptions.lock().await.cancel(cid)` here as well. I don't think we should do it right away since we don't have any way to communicate to bitswap we don't actually need the block after all.

http/Cargo.toml
```diff
@@ -28,3 +28,6 @@ structopt = "0.3.12"
 thiserror = "1.0.14"
 tokio = { version = "0.2.16", features = ["full"] }
 warp = "0.2.2"
+url = "2.1.1"
+async-stream = "0.2.1"
```
koivunej, line 32: I don't think this is used anywhere.

http/src/v0/block.rs
```diff
@@ -108,31 +112,176 @@ pub fn put<T: IpfsTypes>(
...
+            hash: "".to_string(),
+            error: e.to_string(),
+        }),
+    }
+}
+
+async fn rm_blocks<T: IpfsTypes>(
+    ipfs: Ipfs<T>,
+    cids: Vec<Cid>,
+) -> Result<impl Stream<Item = Result<RmResponse, Error>> + Send, Error> {
```
koivunej, line 240:
```suggestion
) -> FuturesOrdered<Item = Result<RmResponse, Error>> {
```
The result seems extra, as I can't see a way for this to error completly?

http/src/v0/block.rs
```diff
@@ -108,31 +112,176 @@ pub fn put<T: IpfsTypes>(
...
+        .into_iter()
+        .map(|hash| Cid::from_str(&hash).map_err(StringError::from).unwrap())
+        .collect();
+
+    println!("{:?}", cids);
+
+    let st = rm_blocks(ipfs, cids)
+        .await
+        .map_err(StringError::from)?
+        .map(|result| result.unwrap())
```
koivunej, line 266: This result would go away if you removed the unnecessary Result from around rm_blocks

http/src/v0/block.rs
```diff
@@ -108,31 +112,176 @@ pub fn put<T: IpfsTypes>(
...
+) -> Result<impl Reply, Rejection> {
+    let cids: Vec<Cid> = options
+        .args
+        .into_iter()
+        .map(|hash| Cid::from_str(&hash).map_err(StringError::from).unwrap())
+        .collect();
+
+    println!("{:?}", cids);
+
+    let st = rm_blocks(ipfs, cids)
```
koivunej, line 263: I think if you'd inline the `rm_blocks` and `rm_block` first here before attempting any kind of nice refactoring, this might come out cleaner.

http/src/v0/block.rs
```diff
@@ -108,31 +112,176 @@ pub fn put<T: IpfsTypes>(
...
+        .map(|result| match result {
+            Ok(mut string) => {
+                string.push('\n');
+                Ok(string.into_bytes())
+            }
+            Err(e) => {
+                log::error!("edge serialization failed: {}", e);
+                Err(HandledErr)
+            }
+        });
```
koivunej, line 277: Might be worth while to try to later push these back to the StreamResponse and make it a "JsonNewLineResponse" since they are now repeated in 3 places I think.

src/repo/fs.rs
aphelionz: This is great! I will definitely think about this and implement something close to what you have proposed here.

http/src/v0/block.rs
aphelionz: I think it's safe to assume, based on @alanshaw 's https://github.com/alanshaw/iterable-ndjson, that any time we're returning ndjon that it's expected as a stream, like in the case of removing a large amount of blocks, but if worse comes to worst I'll switch to a non-streaming response

src/repo/fs.rs
aphelionz: Should we return the Cid instead of ()?

http/src/v0/block.rs
aphelionz: Looks like this is not how you annotate a `FuturesOrdered`, getting the "associated type bindings are not allowed here

http/src/v0/block.rs
```diff
@@ -98,31 +102,169 @@ pub fn put<T: IpfsTypes>(
...
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        use ParseError::*;
+        match *self {
+            DuplicateField(ref s) => write!(fmt, "field {:?} was duplicated", *s),
+            MissingArg => write!(fmt, "required field \"arg\" missing"),
+            InvalidNumber(ref k, ref v) => write!(fmt, "field {:?} invalid number: {:?}", *k, *v),
+            InvalidBoolean(ref k, ref v) => write!(fmt, "field {:?} invalid boolean: {:?}", *k, *v),
+        }
+    }
+}
```
koivunej, line 167: just realized: these should go under `v0::support::option_parsing` or something and just right away implement the `warp::reject::Rejection` or whatever the first trait was. I think currently these go through `StringError` for no reason.

http/src/v0/block.rs
```diff
@@ -98,31 +102,169 @@ pub fn put<T: IpfsTypes>(
...
+        .await
+        .into_iter()
+        .map(|result| match result {
+            Ok(cid) => RmResponse {
+                hash: cid.to_string(),
+                error: "".to_string(),
+            },
+            Err(e) => {
+                RmResponse {
+                    // FIXME: How to get the `cid` here, though?
```
koivunej, line 246: you'll need to carry it next to the future on line [R232](https://github.com/ipfs-rust/rust-ipfs/pull/153/files#diff-ee377a5ddf5a7c29cad06ac34d1512edR232) like `.map(|cid| (cid.clone(), ipfs.remove_block(cid))` and fight a few compiler errors.

http/src/v0/block.rs
```diff
@@ -98,31 +102,169 @@ pub fn put<T: IpfsTypes>(
...
+        .args
+        .into_iter()
+        .map(|hash| Cid::from_str(&hash).map_err(StringError::from).unwrap())
+        .map(|cid| ipfs.remove_block(cid))
+        .collect();
+
+    let st = futs
+        .collect::<Vec<Result<Cid, Error>>>()
+        .await
+        .into_iter()
```
koivunej, line 238: These look extra; it's already a stream so just map over it.

src/repo/results.rs
```diff
@@ -0,0 +1,17 @@
...
+    // TODO: DownloadCancelled(Cid, Duration),
+}
+
+// pub struct BlockNotFound(Cid);
+
+#[derive(Debug)]
+pub enum BlockRmError {
+    // TODO: Pinned(Cid),
+    NotFound(Cid),
+}
```
koivunej, line 17: I don't think it's a good idea to create a new file for this, there's plenty of room in the `src/repo/mod.rs`... In general I dislike grouping by type instead of feature. If that starts to create too large files, perhaps we have too many things which need to live next to each other and so our features need to become smaller :)

http/src/v0/block.rs
aphelionz: nope 
image

http/src/v0/block.rs
aphelionz: I can't put .map_err after remove_block, which is async

koivunej: Not sure why? https://docs.rs/futures/0.3.4/futures/future/trait.TryFutureExt.html#method.map_err

http/src/v0/block.rs
aphelionz: Do you want me to do this now or are you just making a note?

src/lib.rs
```diff
@@ -334,8 +334,10 @@ impl<Types: IpfsTypes> Ipfs<Types> {
     }
 
     /// Remove block from the ipfs repo.
-    pub async fn remove_block(&self, cid: &Cid) -> Result<(), Error> {
-        Ok(self.repo.remove_block(cid).await?)
+    pub async fn remove_block(&self, cid: Cid) -> Result<Cid, Error> {
+        let cid = cid.clone();
+        self.repo.remove_block(&cid).await?;
+        Ok(cid)
```
aphelionz, line 340: Unsure if I should be just returning the cid here instead of passing it all the way down and returning it all the way back up, or some mix of both, or if it makes a big difference. I was also thinking this could take a `Cid` and return a `RemovedCid`

http/src/v0/block.rs
```diff
@@ -156,3 +304,84 @@ pub fn stat<T: IpfsTypes>(
         .and(query::<StatQuery>())
         .and_then(stat_query)
 }
+
+use pin_project::pin_project;
```
aphelionz, line 308: Will use these from the `refs` implementation. At some point these should probably be moved to `support`

http/src/v0/block.rs
koivunej: I think it was you who copypasted so I'd of course prefer not to start a cleanup PR but we are in a similar situation with the IpfsPath which would need to move to ipfs, except that it isn't copypasted.

I don't think you rebased this branch yet? When you do it'll be a good time to remove these duplicates.

src/lib.rs
koivunej: This needs to account for cid upgrade I think, you'll have that available on rebase.

The clone itself again I think redundant, clippy probably warned on that already? I think the incoming cid should be taken as a reference, so removing it isn't such of a big deal. If the caller wants to continue using their now removed cid they can. If we want it for an json rendering use case we can just use our orginal value for that.

http/src/v0/block.rs
koivunej: I think they should be moved right away since they were just copypasted.

http/src/v0/block.rs
```diff
@@ -98,31 +102,174 @@ pub fn put<T: IpfsTypes>(
...
+        match *self {
+            DuplicateField(ref s) => write!(fmt, "field {:?} was duplicated", *s),
+            MissingArg => write!(fmt, "required field \"arg\" missing"),
+            InvalidNumber(ref k, ref v) => write!(fmt, "field {:?} invalid number: {:?}", *k, *v),
+            InvalidBoolean(ref k, ref v) => write!(fmt, "field {:?} invalid boolean: {:?}", *k, *v),
+        }
+    }
+}
+
+impl<'a> TryFrom<&'a str> for RmOptions {
```
koivunej, line 169: could do similar splitting as I did with `refs`, move this to `block/options.rs`

http/src/v0/block.rs
```diff
@@ -98,31 +102,174 @@ pub fn put<T: IpfsTypes>(
...
+    let futs: FuturesOrdered<_> = args
+        .into_iter()
+        .map(|hash| Cid::from_str(&hash).map_err(StringError::from).unwrap())
+        .map(|cid| ipfs.remove_block(cid.clone()).map_err(move |e| (cid, e)))
+        .collect();
+
+    let responses = futs
+        .collect::<Vec<Result<Cid, (Cid, Error)>>>()
+        .await
+        .into_iter()
```
koivunej, line 233: I still don't understand the collecting to a vec here, am I missing something why all the results need to be buffered?

http/src/v0/block.rs
```diff
@@ -98,31 +102,174 @@ pub fn put<T: IpfsTypes>(
...
+                }
+                Ok(string.into_bytes())
+            }
+            Err(e) => {
+                log::error!("edge serialization failed: {}", e);
+                Err(HandledErr)
+            }
+        });
+
+    let st = futures::stream::iter(responses);
```
koivunej, line 269: isn't FuturesOrdered already a stream?

src/lib.rs
koivunej: Aah not here, this was the `Ipfs`, but in the `Repo`

src/repo/mod.rs
```diff
@@ -241,20 +255,34 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
...
-        Ok(())
+
+        // FIXME: Need to change location of pinning logic.
+        // I like this pattern of the repo abstraction being some sort of
+        // "clearing house" for the underlying result enums, but this
+        // could potentially be pushed out out of here up to Ipfs, idk
+        match self.block_store.remove(&cid).await? {
+            Ok(success) => match success {
+                BlockRm::Removed(cid) => Ok(cid),
+                // TODO: Add DownloadCancelled here
```
koivunej, line 279: these todos can be removed as if you go and add a variant, this will start giving out a compiler error

src/repo/mod.rs
```diff
@@ -241,20 +255,34 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
...
-        self.block_store.remove(&cid).await?;
-        Ok(())
+
+        // FIXME: Need to change location of pinning logic.
+        // I like this pattern of the repo abstraction being some sort of
+        // "clearing house" for the underlying result enums, but this
+        // could potentially be pushed out out of here up to Ipfs, idk
+        match self.block_store.remove(&cid).await? {
+            Ok(success) => match success {
+                BlockRm::Removed(cid) => Ok(cid),
```
koivunej, line 278: the issue here is that all of the methods I touched while doing the cidv0->cidv1 is that if they return the cid, they return the cid the user used, not the upgraded one as is done here.

aphelionz: I removed the clone from Ipfs::remove_block, which will pass the user-supplied Cid to Repo::remove_block, which upgrades it and then passes it down to the BlockStore That should be sufficient, right? We could potentially upgrade it after retrieval

http/src/v0/block.rs
koivunej: Tried this, got a `cannot infer an appropriate lifetime for autoref due to conflicting requirements` ... Probably because there's nothing keeping the `ipfs` alive until the end.

src/repo/mod.rs
koivunej: I think on this line it should be optionally downgraded, which is a bit of chore (you'd need to check version, codec and hash). I think the easiest would be to rename the original parameter as `original`, keep the other cids but return `Ok(original.clone())` here.

It does however make the `BlockRm::Removed(HERE)` <-- the HERE a bit redundant.

koivunej: Another way is to just remove the `Cid` from return value and use the original `src/lib.rs` version here: https://github.com/ipfs-rust/rust-ipfs/pull/153/files#diff-b4aea3e418ccdb71239b96952d9cddb6R337.

If the cid is not taken by value, the caller already has the original id and can continue using it.

PR #160, opened 2020-04-17
fix: minor block/rm bug

aphelionz:
This PR fixes two minor bugs found while doing final 1.2 conformance testing:
1. ~Refs local wants the results to be returned as `Qm***` multihashes~
2. Block rm had a minor typo-related bug, switched two lines.

Edit: only the block rm fix now


aphelionz:
Need to look at the failing test after I've had some sleep :)

PR #182, opened 2020-06-08
Update README.md

aphelionz:


koivunej:
Almost forgot: https://github.com/rs-ipfs/rust-ipfs/pull/182/files#diff-04c6e90faac2675aa89e2176d2eec7d8R113 this needed to be `/shutdown`. I'll add other fixes if there are any once you edit that section and I can suggest.

PR #188, opened 2020-06-12
Remove the / before target can help ignore build files

kevingzhang:
Besides /target, there are many target folders inside examples. Please change /target to target so that those build files can be ignored by github

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-/target
+target
 **/*.rs.bk


koivunej:
Hey thanks @kevingzhang! I had actually previously proposed `**/target` but this does the same but with less.

> Besides /target, there are many target folders inside examples.

Are you getting more than `examples/target` on your build?

The `target` dirs I have on my system (still using RLS for some reason):

```
$ find . -type d -and -name 'target'
./http/target
./bitswap/target
./target
./unixfs/target
./examples/target
```

PR #204, opened 2020-06-25
make examples cargo run-nable again

ljedrz:
Users usually expect to be able to run those from the root directory of the crate, so enable that. This also removes the unused extra `main.rs`.


aphelionz:
I don't believe the examples are run as part of CI, can we add those?

ljedrz:
@aphelionz they are compiled; due to some of them requiring user input I think we are good as-is.

Issue #178, opened 2020-06-08
No latency in verbose swarm/peers before a ping response


koivunej:
Test log: https://github.com/rs-ipfs/rust-ipfs/pull/177/checks?check_run_id=749321323#step:9:192

I am not sure what the other implementations could possibly be doing here. Counting up from the connection established would be the correct way to go about this.

_Originally posted by @koivunej in https://github.com/rs-ipfs/rust-ipfs/pull/177#issuecomment-640487950_

koivunej:
Added the good first issue label and here's some initial thinking I've had for possible implementation. The current implementation just [records the latest ping roundtrip time](https://github.com/rs-ipfs/rust-ipfs/blob/master/src/p2p/swarm.rs#L85-L88) when they arrive. This is later [rendered into the json at corresponding swarm http handler](https://github.com/rs-ipfs/rust-ipfs/blob/master/http/src/v0/swarm.rs#L61-L65).

Possible steps to take here would be something like: 

* optional: peek at go/js-ipfs impls for what they are doing in a such case
* in ipfs::p2p::swarm, [record an instant when a peer was connected](https://github.com/rs-ipfs/rust-ipfs/blob/master/src/p2p/swarm.rs#L161), then return that [`Instant::elapsed(&self)`](https://doc.rust-lang.org/stable/std/time/struct.Instant.html#method.elapsed) [as the RTT](https://github.com/rs-ipfs/rust-ipfs/blob/master/src/p2p/swarm.rs#L71)

koivunej:
In #212 the CI test failures were removed by returning "n/a" when no latency was known, following `js-ipfs` implementation. The implementation suggestion at https://github.com/rs-ipfs/rust-ipfs/issues/178#issuecomment-640638258 is still valid though.

PR #210, opened 2020-06-26
A round of cleanups

ljedrz:
Assorted drive-by cleanups and small refactorings; individual commits describe the specific changes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -476,11 +476,11 @@ impl<Types: IpfsTypes> Ipfs<Types> {
 
         self.to_task
             .clone()
-            .send(IpfsEvent::PubsubSubscribe(topic, tx))
+            .send(IpfsEvent::PubsubSubscribe(topic.clone(), tx))
             .await?;
 
         rx.await?
-            .ok_or_else(|| format_err!("already subscribed to the given topic"))
+            .ok_or_else(|| format_err!("already subscribed to {:?}", topic))
     }
 
     /// Publishes to the topic which may have been subscribed to earlier
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -471,25 +471,25 @@ impl<Types: IpfsTypes> Ipfs<Types> {
     /// Subscribes to a given topic. Can be done at most once without unsubscribing in the between.
     /// The subscription can be unsubscribed by dropping the stream or calling
     /// [`pubsub_unsubscribe`].
-    pub async fn pubsub_subscribe(&self, topic: &str) -> Result<SubscriptionStream, Error> {
+    pub async fn pubsub_subscribe(&self, topic: String) -> Result<SubscriptionStream, Error> {
         let (tx, rx) = oneshot_channel();
 
         self.to_task
             .clone()
-            .send(IpfsEvent::PubsubSubscribe(topic.into(), tx))
+            .send(IpfsEvent::PubsubSubscribe(topic, tx))
             .await?;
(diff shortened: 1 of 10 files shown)


ljedrz:
The CI failure seems to be https://github.com/rs-ipfs/rust-ipfs/issues/178.

koivunej:
Would removal of those warnings from latest nightly be good enough to include in cleanups? The annotations are at the bottom of the review. Basically just delete the lines `#[allow(rust_2018_idioms)]`, there should be two of these.

ljedrz:
Ah, I just thought we didn't mind, since that file was auto-generated; btw, since it is, do we even need to keep it in the repo?

koivunej:
> Ah, I just thought we didn't mind, since that file was auto-generated; btw, since it is, do we even need to keep it in the repo?

quick-protobuf doesn't (yet) have nice automation for this, so, so far we do need to keep it. It is also nice to be able to view it. It's not like we should need to rebuild the file anytime soon. But when we do, we should change the field names, which I didn't modify the first time around.

src/lib.rs
```diff
@@ -471,25 +471,25 @@ impl<Types: IpfsTypes> Ipfs<Types> {
     /// Subscribes to a given topic. Can be done at most once without unsubscribing in the between.
     /// The subscription can be unsubscribed by dropping the stream or calling
     /// [`pubsub_unsubscribe`].
-    pub async fn pubsub_subscribe(&self, topic: &str) -> Result<SubscriptionStream, Error> {
+    pub async fn pubsub_subscribe(&self, topic: String) -> Result<SubscriptionStream, Error> {
```
koivunej, line 474: So basically the rule here for the `ipfs::Ipfs` API would be: take ownership of the value we know we'll have to have clone either way.

src/lib.rs
```diff
@@ -471,25 +471,25 @@ impl<Types: IpfsTypes> Ipfs<Types> {
...
 
         self.to_task
             .clone()
-            .send(IpfsEvent::PubsubSubscribe(topic.into(), tx))
+            .send(IpfsEvent::PubsubSubscribe(topic, tx))
             .await?;
 
         rx.await?
-            .ok_or_else(|| format_err!("already subscribed to {:?}", topic))
+            .ok_or_else(|| format_err!("already subscribed to the given topic"))
```
koivunej, line 483: I usually do error messages in `constant_part: variable_part` or "like [perror(3)](https://linux.die.net/man/3/perror)" but here I failed with it apparently.

I am still very much in favor of including more the topic in this error for example. Errors like this are quite dreadful to have happen in production, and then even the stacktrace doesn't help at all, since the error comes from the internal state which the user may have been modifying concurrently by dropping the previous receiver, but it all happens in quite unexplainable order because of the many queues in play.

src/lib.rs
ljedrz: I believe it is preferable to be upfront about the need to own the value when possible, instead of having concealed clones after providing a reference. It is unlikely to be any issue in this particular case, just a general rule that can't hurt.

src/lib.rs
koivunej: More examples for "Context for errors" can be found related to this: https://github.com/dvc94ch/rust-ipld/pull/1 which was adding context to quite strange error there was for a incorrectly formatted Cid (`b"0".concat(cid.to_bytes())`), it was a bit tricky to track down.

PR #216, opened 2020-06-30
A handful of perf/brevity bits and bobs

ljedrz:
Individual commits describe specific changes.

### DIFF
--- a/http/src/v0/block.rs
+++ b/http/src/v0/block.rs
@@ -163,19 +163,10 @@ async fn rm_query<T: IpfsTypes>(
                 hash: cid.to_string(),
                 error: "".to_string(),
             },
-            Err((cid, e)) => {
-                if force {
-                    RmResponse {
-                        hash: cid.to_string(),
-                        error: "".to_string(),
-                    }
-                } else {
-                    RmResponse {
-                        hash: cid.to_string(),
-                        error: e.to_string(),
-                    }
-                }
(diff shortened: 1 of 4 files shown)


http/src/v0/refs/path.rs
```diff
@@ -133,7 +133,7 @@ impl IpfsPath {
                 // FIXME: this would require the iterator to be peekable in addition.
                 return Ok(WalkSuccess::Link(key, cid));
             }
-            Ipld::Map(mut m) if m.contains_key(&key) => {
+            Ipld::Map(mut m) => {
```
koivunej, line 136: Aah of course perhaps I refactored this back and forth and forgot. I don't think error is ever checked. Btw, here the error sucks, moving stuff into the error is ... too much, no-one probably ever needs the partial documents and those might be too large to print either way.

http/src/v0/block.rs
```diff
@@ -163,19 +163,10 @@ async fn rm_query<T: IpfsTypes>(
...
-                    RmResponse {
-                        hash: cid.to_string(),
-                        error: e.to_string(),
-                    }
-                }
-            }
+            Err((cid, e)) => RmResponse {
+                hash: cid.to_string(),
+                error: if force { "".to_string() } else { e.to_string() },
+            },
```
koivunej, line 169: good catch!

PR #218, opened 2020-07-01
remove the Strategy generic & replace libipld with cid in bitswap

ljedrz:
- we only support a single strategy and the fact it is a generic complicates things; make it non-generic, which paves way to improving event handling
- `cid` is a smaller dep than `libipld` and we only use its bits in `bitswap`


koivunej:
This failure does look interesting... Everything passed but still failure?

bors retry

PR #221, opened 2020-07-02
remove Strategy, improve event handling

ljedrz:
The scope of changes is far greater than I've anticipated, but bear with me.

Key, planned changes:
- remove `bitswap::strategy`
- move `bitswap::Stats` to `ledger`
- introduce `BitswapEvents` emitted by `Bitswap::poll`
- move `Message` into `Ledger`

Drive-by changes forced by current design:
- wrap `Stats` in an `Arc`, make individual stats `AtomicU64`
- add a `Bitswap.queued_blocks` collection populated from `IpfsFuture::poll`
- make some child `bitswap` objects `pub` or `pub(crate)`
- update `Stats` from `IpfsFuture::poll` (good causality-wise, bad encapsulation-wise)
- introduce a `Bitswap.io_events` sender and its `IpfsFuture.bitswap_events` counterpart receiver

Observations:
- `Repo` is something we want to have access to both in `bitswap` (in order to find `want`ed `Block`s) and in `Ipfs` (to handle the higher-level API requests)
- `Ledger`s for the different `PeerId`s contain `Stats` that are updated at `IpfsFuture` level
- `BitswapEvents` don't currently result in anything of interest, due to `Behaviour` not containing the `Repo`

Key takeaway: we need to do some re-designing in order for key objects to have better access to one another.


src/lib.rs
```diff
@@ -876,6 +879,72 @@ impl Future for IpfsFuture {
...
+                            "Peer {} wants block {} with priority {}",
+                            peer_id.to_base58(),
+                            cid.to_string(),
+                            priority
+                        );
+
+                        let repo = self.repo.clone();
+                        let queued_blocks = Arc::clone(&self.swarm.bitswap().queued_blocks);
+                        task::spawn(async move {
+                            let block = match repo.get_block(&cid).await {
```
koivunej, line 928: Unrelated to anything introduced in this PR, perhaps a byproduct of the changes is that using `get_block` directly will put to block on our wantlist. If we had working discovery, we would be gathering wantlists from every peer we meet and thus being highly helpful in the network while running out of memory to help everyone we've met :)

```suggestion
                            // FIXME: need a better API to `get_block_now` or without waiting.
                            // use of `get_block` here results in us trying wanting the block. while
                            // such "helping a friend out" might be useful that'd be an highly
                            // advanced feature we probably don't yet want to pursue.
                            let block = match repo.get_block(&cid).await {
```

koivunej: this will be handled on a follow-up PR with even more changes

Issue #174, opened 2020-05-25
Replace std::sync::Mutex in the subscriptionregistry with one from futures


koivunej:
Direct use of std::sync::Mutex should be avoided.

Labels goodfirstissue and helpwanted: This is a good first issue as this is quite clearly defined fix.
Bug because of deadlock proneness of using std::sync::Mutex in async contexts.

Other such instances can probably be found from the codebase.

koivunej:
Adding labels goodfirstissue and helpwanted as this is a good first issue as this is quite clearly defined fix. Bug is because of deadlock proneness of using std::sync::Mutex in async contexts. Other such instances can probably be found from the codebase.

koivunej:
With #226 and #239 this might be handled as well as it can be in the current design. The solution in the end was to use [block_on](https://docs.rs/async-std/1.5.0/async_std/task/fn.block_on.html) to lock `futures` mutexes in cases where it would had been difficult otherwise.

PR #226, opened 2020-07-07
Assorted improvements to Repo and SubscriptionRegistry

ljedrz:
Individual commits describe the specific changes; notable ones:
- remove the `Arc` and its clones where not needed
- reduce allocations around subscription wakers
- use `futures::lock::Mutex` instead of `std::sync::Mutex` where applicable
- add a `wipe` method to the stores (useful for tests/benches)

cc https://github.com/rs-ipfs/rust-ipfs/issues/174


src/subscription.rs
```diff
@@ -140,18 +136,12 @@ impl<TResult> fmt::Debug for Subscription<TResult> {
...
 impl<TResult> Subscription<TResult> {
-    pub fn new() -> Self {
-        Self {
-            result: Default::default(),
-            wakers: Default::default(),
-            cancelled: false,
+    pub fn add_waker(&mut self, waker: &Waker) {
+        if !self.wakers.iter().any(|w| w.will_wake(waker)) {
+            self.wakers.push(waker.clone());
         }
```
koivunej, line 142: good find! didn't know such existed.

src/subscription.rs
```diff
@@ -9,7 +9,7 @@ use std::sync::{Arc, Mutex};
 
 pub struct SubscriptionRegistry<TReq: Debug + Eq + Hash, TRes: Debug> {
     subscriptions: HashMap<TReq, Arc<Mutex<Subscription<TRes>>>>,
-    cancelled: bool,
+    shutting_down: bool,
 }
 
 impl<TReq: Debug + Eq + Hash, TRes: Debug> fmt::Debug for SubscriptionRegistry<TReq, TRes> {
```
koivunej, line 15: unable to comment on the actual lines (21, 22) but the changes to this file reminded me: the Debug formatter is broken (`std::any::type_name::<Self>()` already includes the TReq and TRes so the output has the two TReq and TRes twice.

ljedrz: I seem to recall that bit; I'll include that with my next changes :+1:

PR #227, opened 2020-07-08
Assorted improvements, continued

ljedrz:
~~Builds on https://github.com/rs-ipfs/rust-ipfs/pull/226, only the last 5 commits are new.~~
The commit messages describe the specific changes, of which the following are the most interesting:
- don't clone the `Subscription`'s result unless it's ready
- enum-ify the `Subscription` object

### DIFF
--- a/bitswap/src/protocol.rs
+++ b/bitswap/src/protocol.rs
@@ -18,8 +18,8 @@ const MAX_BUF_SIZE: usize = 524_288;
 
 type FutureResult<T, E> = Pin<Box<dyn Future<Output = Result<T, E>> + Send>>;
 
-#[derive(Clone, Debug, Default)]
-pub struct BitswapConfig {}
+#[derive(Clone, Copy, Debug, Default)]
+pub struct BitswapConfig;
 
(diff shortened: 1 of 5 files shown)


bitswap/src/protocol.rs
```diff
@@ -18,8 +18,8 @@ const MAX_BUF_SIZE: usize = 524_288;
 
 type FutureResult<T, E> = Pin<Box<dyn Future<Output = Result<T, E>> + Send>>;
 
-#[derive(Clone, Debug, Default)]
-pub struct BitswapConfig {}
+#[derive(Clone, Copy, Debug, Default)]
+pub struct BitswapConfig;
```
koivunej, line 22: Are we sure this will be `Copy` into the future? Wouldn't it be safer to leave this `!Copy`?

ljedrz: it is cheaper for now; we are free to remove it later when this object is extended; adding anything non-`Copy` inside will immediately notify us, so there's no risk of forgetting it

Issue #236, opened 2020-07-10
Block want cancellation


koivunej:
When all futures generated calling `ipfs::Ipfs::get_block` with an equal Cid are dropped, the want for the Cid should disappear. There are now a test cases for this in the conformance tests under [bitswap, which we currently ignore](https://github.com/rs-ipfs/ipfs-rust-conformance/blob/a5675f61ecb472504925d95969880829671691c7/test/index.js#L62-L64):

* 'should remove blocks from the wantlist when requests are cancelled'
* 'should keep blocks in the wantlist when only one request is cancelled'

We've already gone back and forth on the design but a good way of doing this is still hiding.

Regarding the conformance test cases, it would seem that we needlessly hang on /api/v0/dag/resolve?arg=QmSoLPppuBtQSGwKDZT2M73ULpjvfd3aZ6ha4oFGL1KaGa where as go-ipfs returns immediatedly and a /block/get starts, which is properly cancelled. It seems that paths which are empty should be resolved right away.

PR #239, opened 2020-07-10
cancel Blocks from the wantlist when all their Subscriptions die

ljedrz:
I'd have split these changes into more commits, but a lot has changed in the meantime - this was not a trivial thing to do right, that's for sure :dizzy_face:.

~~In any case, it is _almost_ done: the last thing left is to make sure the `Cid` in the wantlist and the subscription list matches, as now we upgrade the latter, which makes the test loop indefinitely if the `Cid` differs between versions.~~ deferred

~~partially solves https://github.com/rs-ipfs/rust-ipfs/issues/236; the functionality is there, but the JS conformance test doesn't pass - might need patching~~
actually closes https://github.com/rs-ipfs/rust-ipfs/issues/236

### DIFF
--- a/src/repo/mod.rs
+++ b/src/repo/mod.rs
@@ -134,7 +134,7 @@ trait CidUpgrade: Sized {
 }
 
 /// Extension trait similar to [`CidUpgrade`] but for non-owned types.
-trait CidUpgradedRef: Sized {
+pub trait CidUpgradedRef: Sized {
     /// Returns the otherwise the equivalent value from `&self` but with Cid version 1
     fn as_upgraded_cid(&self) -> Self;
 }
@@ -221,8 +221,7 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
         self.subscriptions
             .lock()
             .await
-            // subscriptions are per cidv1
-            .finish_subscription(&cid.clone().into(), block);
+            .finish_subscription(&original_cid.clone().into(), block);
         // sending only fails if no one is listening anymore
         // and that is okay with us.
         self.events
@@ -249,8 +248,7 @@ impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
                 .subscriptions
                 .lock()
                 .await
-                // subscribe always by using the cidv1
-                .create_subscription(upgraded.clone().into(), Some(self.events.clone()));
+                .create_subscription(cid.clone().into(), Some(self.events.clone()));
             // sending only fails if no one is listening anymore
             // and that is okay with us.
             self.events
--- a/src/repo/mod.rs
+++ b/src/repo/mod.rs
@@ -1,7 +1,7 @@
 //! IPFS repo
 use crate::error::Error;
 use crate::path::IpfsPath;
-use crate::subscription::SubscriptionRegistry;
+use crate::subscription::{Request, SubscriptionRegistry};
 use crate::IpfsOptions;
 use async_std::path::PathBuf;
 use async_trait::async_trait;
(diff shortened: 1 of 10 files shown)


src/subscription.rs
```diff
@@ -1,19 +1,64 @@
...
-use std::sync::{Arc, Mutex};
+use std::sync::{
+    atomic::{AtomicBool, AtomicU64, Ordering},
+    Arc,
+};
 
-pub struct SubscriptionRegistry<TReq: Debug + Eq + Hash, TRes: Debug> {
-    subscriptions: HashMap<TReq, Arc<Mutex<Subscription<TRes>>>>,
-    shutting_down: bool,
+static GLOBAL_REQ_COUNT: AtomicU64 = AtomicU64::new(0);
```
koivunej, line 19: maybe add a comment what is this used for?

src/repo/mod.rs
```diff
@@ -103,16 +102,31 @@ pub struct Repo<TRepoTypes: RepoTypes> {
...
+            kind: RequestKind::GetBlock(cid),
+            ..
+        } = req
+        {
+            RepoEvent::UnwantBlock(cid)
+        } else {
+            panic!("logic error: RepoEvent can only be created from a Request::GetBlock");
+        }
+    }
+}
```
koivunej, line 128: since this conversion can panic, I wonder if it should be `fn to_repo_event(self) -> Result<RepoEvent, Self>` instead? having the panic message from here would perhaps be more suitable where the conversion is needed inside `.expect(...)`?

Though, I haven't yet found where this is used.

ljedrz: that panic is a logic error, though - it _should_ panic if it's used against its purpose

tests/wantlist_and_cancellation.rs
```diff
@@ -0,0 +1,109 @@
...
+        Duration::from_millis(200),
+        || ipfs.bitswap_wantlist(None),
+        |ret| ret.unwrap().is_empty(),
+    );
+
+    assert!(
+        wantlist_cleared.is_ok(),
+        "a block was not removed from the wantlist after all its subscriptions had died"
+    );
+}
```
koivunej, line 109: Just reminding about going back to the test with just a single task using `futures::future::select` and all the complexity it brings, while unidiomatic rust code, it helped uncover bugs whereas this is more tolerant to such bugs (waker replacement).

src/repo/mod.rs
ljedrz: though it is a `From` impl, so it makes sense that it shouldn't fail; I'll change it to a `TryFrom` and add an `unwrap` to the callsite.

src/subscription.rs
```diff
@@ -112,70 +181,147 @@ impl fmt::Display for Cancelled {
...
+            (sub, is_last)
+        });
+
+        if let Some(sub) = sub {
+            if let Subscription::Pending { ref request, .. } = sub {
+                debug!("Dropping subscription {}", request.id);
+            }
+            // don't bother updating anything that isn't `Pending`
+            if let mut sub @ Subscription::Pending { .. } = sub {
+                debug!("It was the last subscription to a resource, sending a cancel notification");
```
koivunej, line 312:
```suggestion
                debug!("It was the last subscription to a resource, sending a cancel notification: {:?}", sub);
```

maybe?

koivunej: wait couldn't these two matches be combined, maybe combined into the upper, maybe into a single match explaining why the other need not to be cancelled, or something like `Some(_) => { /* nothing to cancel for Cancelled or Ready */ }`?

src/subscription.rs
```diff
@@ -24,76 +95,74 @@ impl<TReq: Debug + Eq + Hash, TRes: Debug> fmt::Debug for SubscriptionRegistry<T
...
-impl<TReq: Debug + Eq + Hash, TRes: Debug> Default for SubscriptionRegistry<TReq, TRes> {
+impl<TRes: Debug + Clone + PartialEq> Default for SubscriptionRegistry<TRes> {
     fn default() -> Self {
         Self {
             subscriptions: Default::default(),
-            shutting_down: false,
+            shutting_down: Default::default(),
         }
     }
 }
```
koivunej, line 163: Cannot this be `#[derive(Default)]` because it adds some `TRes: Default` bounds or is there some other issue I cannot see?

src/subscription.rs
ljedrz: the subscription is already described in the debug right above it, and that one properly handles the ugly display of `Cid` (not a straight `Debug`, but a full `Display` impl)

ljedrz: ah no, I actually already changed it; in that case I'll do it so it does :laughing:

src/subscription.rs
ljedrz: yep, and `TRes` unfortunately can't be `Default`

koivunej:
Well... duh. I already went ahead and merged the ipfs-rust-conformance change. This seems to be stuck in PR CI build, even though it has already been merged in.

bors ping

koivunej:
Tried to cancel the timing out mac os build.

koivunej:
I am pretty sure this PR should had been automatically closed as it's been merged, but the "Checking for ability to merge automatically..." is currently hanging, so perhaps this problem will just resolve itself "later."

koivunej:
Phew :) Finally with the merge status.

Issue #194, opened 2020-06-17
flaky test: listening_for_multiple_unspecified_addresses


koivunej:
readme-doctest failed on a new flaky test which I had not realized earlier was flaky: `listening_for_multiple_unspecified_addresses` in https://github.com/rs-ipfs/rust-ipfs/pull/192/checks?check_run_id=782193837#step:5:379.

Not sure what do with this, but it's clear the case is similar to `multiple_concurrent_ephemeral_listening_addresses_on_same_ip` on #177.

koivunej:
Fixing this now in the same way as previously. Not closing the issue as the general problem with trying to do it this way still remains.

ljedrz:
I can't reproduce the issue neither with `listening_for_multiple_unspecified_addresses` nor `multiple_concurrent_ephemeral_listening_addresses_on_same_ip` when running the test in a loop.

koivunej:
You have too many cores :) I tried to explain the ordering which is happening in the description. I assume you agree on it's validity?

We could get ton of these errors to pop up if we were able to delay a few operations...

ljedrz:
Yeah, just reporting my experience with these tests; FWIW, I also don't seem to have issues under `#[tokio::test(max_threads = 1)]`.

PR #244, opened 2020-07-16
optimize the SubsctiptionRegistry

ljedrz:
Removes the previous linear search in `finish_subscription` and the now-unneeded `Request` object.

I'm pretty sure there is a little bit of `RequestKind` cloning that can be removed, but that's minor in comparison and can be improved upon later.

### DIFF
--- a/src/repo/mod.rs
+++ b/src/repo/mod.rs
@@ -1,7 +1,7 @@
 //! IPFS repo
 use crate::error::Error;
 use crate::path::IpfsPath;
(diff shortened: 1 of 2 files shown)


koivunej:
The failure sounds like https://github.com/rs-ipfs/rust-ipfs/issues/194 ... Might as well just quickly handle it so that it doesnt bother all of the time.

koivunej:
Raah, again the windows builds hangs on conformance tests without any console output.

bors r-

PR #250, opened 2020-07-17
Migrating http symlink creation to conformance/setup.sh + CI Update

aphelionz:
This PR removes the symlink from source control, and adds some lines to validate and create the symlink to `conformance/setup.sh`. It also updates CI to build the project _before_ trying to create the symlink, as per the warning in `setup.sh`

The reasoning behind this is for GitHub Pages + Jekyll, which will break upon encountering a symlink that doesn't resolve properly in the repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,15 +109,15 @@ jobs:
       if: contains(matrix.platform.target, 'android')
       run: cargo install cargo-ndk
 
+    - name: Build (others)
+      if: matrix.platform.cross == false
+      run: cargo build --locked --workspace --all-targets
+
     - name: Setup conformance tests (non-cross targets)
       if: matrix.platform.cross == false
       run: ./setup.sh
       shell: bash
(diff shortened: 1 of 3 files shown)


koivunej:
This has to be rebased to get a successful CI run at minimum.

.github/workflows/ci.yml
```diff
@@ -108,16 +108,16 @@ jobs:
...
     - name: Setup conformance tests (non-cross targets)
       if: matrix.platform.cross == false
       run: ./setup.sh
       shell: bash
       working-directory: ./conformance
 
-    - name: Build (others)
-      if: matrix.platform.cross == false
-      run: cargo build --locked --workspace --all-targets
-
```
koivunej, line 120: Why did this move?

koivunej: Aah, didn't understand what the check did. Oki good. I'll resolve.

koivunej:
I think this is looking good, needs the rebase.

PR #252, opened 2020-07-20
introduce a local get_block_now functionality

ljedrz:
This allows us not to call the `async` counterpart (that spawns a potentially long-lasting `SubscriptionFuture`) when being asked for a `Cid` by other peers.

In addition, expand the `wantlist_cancellation` test a little bit - these changes don't affect it after all, but the extra checks are useful.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -617,12 +617,6 @@ impl<Types: IpfsTypes> Ipfs<Types> {
         // ignoring the error because it'd mean that the background task had already been dropped
         let _ = self.to_task.clone().try_send(IpfsEvent::Exit);
     }
-
-    /// Obtain the locked collection of all the subscriptions for blocks.
-    #[doc(hidden)]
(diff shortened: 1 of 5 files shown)


koivunej:
the readme-doctest failed with:

```
---- wantlist_cancellation stdout ----
thread 'wantlist_cancellation' panicked at 'assertion failed: `(left == right)`
  left: `Some(3)`,
 right: `Some(2)`', tests/wantlist_and_cancellation.rs:45:5


failures:
    wantlist_cancellation
```

Did you look at the skipped case here: https://github.com/rs-ipfs/rust-ipfs/blob/master/conformance/test/index.js#L48?

ljedrz:
Yeah, I'm trying to find the reason for that failure, can't reproduce it at all :thinking:. Ah yeah, I have that other conformance test in my TODO, I can look into it too, maybe it'll help me find the issue.

PR #253, opened 2020-07-21
update some dependencies

ljedrz:
We have been behind in terms of `cid` and `multihash` for a while now, but with the recent update of `libp2p` (as long as we want to follow it closely), those need to be updated too. The biggest obstacle to updating our dependencies is currently the pinned revision of `libipld` and while it'll take more work to adapt our code to its recent upstream version, that revision can easily be forked with updated dependencies, enabling other updates.


koivunej:
hmm that exchange_block failure is interesting, must be the new async-std? lets see if that repeats. Otherwise this is looking really good!

ljedrz:
`async-std` updated, let's see now

ljedrz:
Cc https://github.com/rs-ipfs/rust-ipfs/issues/75

Issue #10, opened 2019-02-22
Locate providers via Kademlia DHT


dvc94ch:

koivunej:
To visualize the "cid correlation" mentioned in the latest edit, think of `cat` operation on a `$root`. In the case of multiblock files, it'll find N links and proceed to `get_block` the first range-matching `links[c]`. Now there is a high likelyhood of any provider of `$root` to also have `$links[c]` but a decreasing likelyhood of anyone having *just* the subtree starting at `$links[c]`. This decreasing likelyhood is related to at minimum the current setting of go-ipfs which allows only creating provider records for the roots (or even just the pinned roots).

PR #256, opened 2020-07-21
some groundwork for peer/content discovery

ljedrz:
Some groundwork for peer and content discovery using `libp2p::kad`.

~~Very much a WIP with plenty of prototyping and wild `println!`s.~~

Cc https://github.com/rs-ipfs/rust-ipfs/issues/10

### DIFF
--- a/http/src/main.rs
+++ b/http/src/main.rs
@@ -134,8 +134,13 @@ fn main() {
     let mut rt = tokio::runtime::Runtime::new().expect("Failed to create event loop");
 
     rt.block_on(async move {
-        let opts: IpfsOptions<ipfs::TestTypes> =
-            IpfsOptions::new(home.clone().into(), keypair, Vec::new(), false);
+        let opts: IpfsOptions<ipfs::TestTypes> = IpfsOptions::new(
+            home.clone().into(),
+            keypair,
+            Vec::new(),
+            false,
+            Some("/ipfs/kad/1.0.0".into()),
+        );
 
         let (ipfs, task) = UninitializedIpfs::new(opts)
             .await
--- a/http/src/main.rs
+++ b/http/src/main.rs
@@ -134,13 +134,8 @@ fn main() {
     let mut rt = tokio::runtime::Runtime::new().expect("Failed to create event loop");
 
     rt.block_on(async move {
-        let opts: IpfsOptions<ipfs::TestTypes> = IpfsOptions::new(
-            home.clone().into(),
-            keypair,
-            Vec::new(),
-            false,
-            Some("/ipfs/kad/1.0.0".into()),
-        );
+        let opts: IpfsOptions<ipfs::TestTypes> =
+            IpfsOptions::new(home.clone().into(), keypair, Vec::new(), false, None);
 
         let (ipfs, task) = UninitializedIpfs::new(opts)
             .await
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,7 +8,7 @@
 #[macro_use]
 extern crate log;
 
-use anyhow::format_err;
+use anyhow::{anyhow, format_err};
 use async_std::path::PathBuf;
 pub use bitswap::{BitswapEvent, Block, Stats};
 pub use cid::Cid;
@@ -91,6 +91,8 @@ pub struct IpfsOptions<Types: IpfsTypes> {
     pub bootstrap: Vec<(Multiaddr, PeerId)>,
     /// Enables mdns for peer discovery when true.
     pub mdns: bool,
+    /// Custom Kademlia protocol name.
+    pub kad_protocol: Option<String>,
 }
 
 impl<Types: IpfsTypes> fmt::Debug for IpfsOptions<Types> {
@@ -102,18 +104,20 @@ impl<Types: IpfsTypes> fmt::Debug for IpfsOptions<Types> {
             .field("bootstrap", &self.bootstrap)
             .field("keypair", &DebuggableKeypair(&self.keypair))
             .field("mdns", &self.mdns)
+            .field("kad_protocol", &self.kad_protocol)
             .finish()
     }
 }
 
 impl IpfsOptions<TestTypes> {
     /// Creates an inmemory store backed node for tests
-    pub fn inmemory_with_generated_keys(mdns: bool) -> Self {
+    pub fn inmemory_with_generated_keys(mdns: bool, kad_protocol: Option<String>) -> Self {
         Self::new(
             std::env::temp_dir().into(),
             Keypair::generate_ed25519(),
             vec![],
             mdns,
+            kad_protocol,
         )
     }
 }
@@ -147,6 +151,7 @@ impl<Types: IpfsTypes> IpfsOptions<Types> {
         keypair: Keypair,
         bootstrap: Vec<(Multiaddr, PeerId)>,
         mdns: bool,
+        kad_protocol: Option<String>,
     ) -> Self {
         Self {
             _marker: PhantomData,
@@ -154,6 +159,7 @@ impl<Types: IpfsTypes> IpfsOptions<Types> {
             keypair,
             bootstrap,
             mdns,
+            kad_protocol,
         }
     }
 }
@@ -203,6 +209,7 @@ impl<T: IpfsTypes> Default for IpfsOptions<T> {
             keypair,
             bootstrap,
             mdns: true,
+            kad_protocol: Some("/ipfs/kad/1.0.0".into()),
         }
     }
 }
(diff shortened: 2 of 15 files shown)


src/lib.rs
```diff
@@ -91,6 +91,8 @@ pub struct IpfsOptions<Types: IpfsTypes> {
     pub bootstrap: Vec<(Multiaddr, PeerId)>,
     /// Enables mdns for peer discovery when true.
     pub mdns: bool,
+    /// Custom Kademlia protocol name.
+    pub kad_protocol: Option<String>,
```
koivunej, line 95: could this be left out for now at least? Probably later on there should be custom config for which protocol ids to enable, but probably not now (to avoid "/lan" and !"/lan" confusion found in the PR already)

src/lib.rs
```diff
@@ -608,6 +616,35 @@ impl<Types: IpfsTypes> Ipfs<Types> {
         rx.await?
     }
 
+    pub async fn bootstrap(&self) -> Result<(), Error> {
```
koivunej, line 619:
```suggestion
    // Initiate a query for random key to discover peers.
    pub async fn bootstrap(&self) -> Result<(), Error> {
```

Is there a reason for this to be in the API or is it only while we are not doing this periodically ourselves?

src/lib.rs
ljedrz: perhaps I can remove the values in the calls, but leave the option itself? seems innocuous and we're likely to want it later on

http/src/main.rs
```diff
@@ -134,8 +134,13 @@ fn main() {
...
 
     rt.block_on(async move {
-        let opts: IpfsOptions<ipfs::TestTypes> =
-            IpfsOptions::new(home.clone().into(), keypair, Vec::new(), false);
+        let opts: IpfsOptions<ipfs::TestTypes> = IpfsOptions::new(
+            home.clone().into(),
+            keypair,
+            Vec::new(),
+            false,
+            Some("/ipfs/kad/1.0.0".into()),
```
koivunej, line 142: elsewhere with "/lan", but recommended not adding this at all: https://github.com/rs-ipfs/rust-ipfs/pull/256/files#r458760223

src/lib.rs
ljedrz: I don't know if we are doing it at all automatically; I can check

src/p2p/behaviour.rs
```diff
@@ -305,6 +452,23 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
+            Ok(id) => Ok(self.kad_subscriptions.create_subscription(id.into(), None)),
+            Err(e) => {
+                error!("kad: can't bootstrap the node: {:?}", e);
+                Err(anyhow!("kad: can't bootstrap the node: {:?}", e))
+            }
+        }
+    }
+
+    pub fn get_closest_peers(&mut self, id: PeerId) -> SubscriptionFuture<Result<(), String>> {
+        let id = id.to_base58();
```
koivunej, line 467: feels strange that the peerid multihash would be converted to base58 ... is this correct?

src/lib.rs
ljedrz: doesn't seem that way, at least not in tests

src/p2p/behaviour.rs
ljedrz: yes, it matches the ids of the peers; what I'm less sure about it the one for block keys, but this one should be fine

http/src/main.rs
ljedrz: I added `/lan` only in test scenarios; as mentioned before, seems I can leave it out for now

src/lib.rs
```diff
@@ -608,6 +616,35 @@ impl<Types: IpfsTypes> Ipfs<Types> {
...
+        let self_peer = PeerId::from_public_key(self.identity().await?.0);
+        let (tx, rx) = oneshot_channel::<FutureSubscription<(), String>>();
+
+        self.to_task
+            .clone()
+            .send(IpfsEvent::GetClosestPeers(self_peer, tx))
+            .await?;
+
+        rx.await?.await?.map_err(|e| anyhow!(e))
+    }
```
koivunej, line 646: I already got confused, should this be the bootstrap or not?

ljedrz: finding closest peers doesn't result in connecting to them at this moment; it's practical for testing DHT functionalities or doing related measurements, not unlike `bitswap_stats` that serves similar purposes

src/lib.rs
koivunej: Yeah definitely no need give this in the inmemory_with_generated_keys since it's not needed in the existing tests at least.

Could you elaborate why we'd like to have this later on?

src/lib.rs
koivunej: No we are not yet doing it automatically, but we probably should eventually.

What is the reason for exposing this as a method? Does kad queue these bootstraps or are they executed concurrently?

src/lib.rs
ljedrz: so we can test both protocols; that recent go-ipfs article suggested they are both significant

src/lib.rs
ljedrz: > Does kad queue these bootstraps or are they executed concurrently?

good question; I don't know, but I'd expect it to be concurrent; a single call can result in a few of logs that the bootstrapping query has completed

src/lib.rs
koivunej: > recent go-ipfs article suggested they are both significant

I think that is for multiple DHTs, not either one of those: lan for private networks and the !lan for the global. I think we should be fine with just the global if anything at the moment.
From above:

>> Yeah definitely no need give this in the inmemory_with_generated_keys since it's not needed in the existing tests at least.

Do you agree we should delete the parameter from `inmemory_with_generated_keys`? I see no value in configuring it in all test scenarios, maybe in some of the dht specific, or if we want to use something specific which is sure to be internal for our test case.

ljedrz: >Do you agree we should delete the parameter from inmemory_with_generated_keys?

yeah, I just meant leaving it in `IpfsOptions`

src/lib.rs
koivunej: How does this allow measuring? bitswap_stats is on the http API (can be used for monitoring) and is interesting when we can start caring about duplicate blocks and such.

src/p2p/behaviour.rs
```diff
@@ -246,7 +384,16 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
+    pub fn add_peer(&mut self, peer: PeerId, addr: Multiaddr) {
+        self.kademlia.add_address(&peer, addr);
+        /*
+                match self.kademlia.start_providing(peer.to_base58().into_bytes().into()) {
+                    Ok(id) => {
+                        self.kad_queries.insert(id);
+                    },
+                    Err(e) => error!("kad: can't provide peer {}: {:?}", peer, e),
+                }
+        */
```
koivunej, line 396: This commented out code seems a bit odd. I would had expected the routing tables would be used without explicitly providing anything for a key. Would something be stored under the other peers key?

src/lib.rs
ljedrz: for measuring how fast we can find discover peers and in what scenarios, e.g. wrt the number of nodes and how they are connected (meshed/linked); I can remove that API, but this was my primary tool to verify that I had peer discovery working :stuck_out_tongue:

src/p2p/behaviour.rs
ljedrz: this should be removed, I forgot to do that; I initially didn't know if it was a necessary step

src/lib.rs
koivunej: Yeah maybe perhaps there could be a method comment to avoid anyone trying to use this for anything, or noting it's used in the test to produce log messages?

src/lib.rs
```diff
@@ -608,6 +616,36 @@ impl<Types: IpfsTypes> Ipfs<Types> {
...
+    // Initiate a query for random key to discover peers.
+    pub async fn bootstrap(&self) -> Result<(), Error> {
+        let (tx, rx) = oneshot_channel::<Result<FutureSubscription<(), String>, Error>>();
+
+        self.to_task.clone().send(IpfsEvent::Bootstrap(tx)).await?;
+
+        rx.await??.await?.map_err(|e| anyhow!(e))
+    }
+
+    pub async fn add_peer(&self, peer_id: PeerId, addr: Multiaddr) -> Result<(), Error> {
```
koivunej, line 628: Could this have a method comment? This is basically to add a known or bootstrap node at runtime?

ljedrz: yeah, will do - and yes, this adds a known node to Kademlia at runtime

src/lib.rs
ljedrz: I can just move it to `Node` that is only used in tests :ok_hand:

koivunej:
I dont think anything changed with the bitswap so that's probably a transient one.

bors retry

PR #259, opened 2020-07-23
Multihash-based repo

ljedrz:
We currently have an ad-hoc `Cid`-upgrading mechanism in place for the legacy `v0` version that is complicated and error-prone; when working on the `SubscriptionRegistry` I had to disable some of those upgrades, but some of them were still applied in the `Repo` (and necessary in order for the conformance tests to pass), making it hard to work on further functionalities like content discovery.

The solution I came up with is to introduce a `RepoCid` wrapper in order to make the keys in `Repo` objects operate on the associated `Multihash` instead of the whole `Cid`. Changing them to plain `Multihash`es wouldn't work, as we are still expected to return `Cid`s for the purposes of e.g. `Ipfs::local_refs` and I'm pretty sure some of the conformance tests still expect the full `Cid` to be available.

The only drawback is some extra `clone()`ing, but the improvement in code clarity and the unblocking of further changes is well worth it. This change also simplifies one of our conformance test patches (local refs).

~~I'm marking this PR as a draft in hopes that I can still come up with some solution to the extra cloning.~~ This is hard; I'm not able to do this at the moment, but this shouldn't be a blocker.


ljedrz:
The only failure is a pre-existing random timeout.

Issue #134, opened 2020-04-02
Flaky multi-node tests


koivunej:
Running the pubsub tests repeatedly like:

```
while target/debug/deps/pubsub-c48a8707038c37a4 publish_between_two_nodes; do echo; echo '...'; echo; done
```

(Your executable name will vary, use `cargo test -- publish_between_two_nodes` to find out)

With all logging at level at `trace` this does produce at least two kinds of failures. Seen errors:

 * `[2020-04-02T09:41:35Z DEBUG libp2p_tcp] Dropped TCP connection to undeterminate peer` when the 10s timeout started
 * `[2020-04-02T09:41:08Z DEBUG libp2p_secio] error during secio handshake IoError(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })` even after sending pubsub rpc's back and forth but before the 10s timeout started

In a highly unscientific "lets run these in multiple shells until they fail, then inspect reason" both seem to happen as often. I think I'll just reference this comment in a tracking issue, this might already be better handled with the just released 0.17 of libp2p. Being resilient on disconnects in the test is one possibility but I am used to running tests in Linux and Windows over loopback without much of an issue. Then again, my systems have probably never been as loaded as any public CI server.

By running other tests than the `pubsub_between_two_nodes` they can also fail but I didn't recover any information on those as 5 tests were executed in parallel and only one of them turned logging on so let it be a small data point.

_Originally posted by @koivunej in https://github.com/ipfs-rust/rust-ipfs/pull/133#issuecomment-607744027_

koivunej:
For some reason the tests have started getting better. This might be because of libp2p 0.19 or just random luck with the other builds running on gha.

ljedrz:
Currently the `exchange_block` test likes to fail; I'll look into it.

ljedrz:
I've found a reliable way of reproducing the issue; I'm suspecting insufficient node cleanup to be the root cause.

ljedrz:
Ok, so it might require a bit of a broader solution, as even though my fix seems to be doing the trick, it sometimes causes some `async` deadlocking, which is not ideal either :smile:.

koivunej:
These have not been flaky at all recently, nor sure is that because of GHA architecture or #307 for example. Lets reopen we get more of these.

PR #260, opened 2020-07-23
Initial content discovery

ljedrz:
An initial implementation of content discovery via DHT, where a `get_block` request that doesn't succeed locally results in a call to `Kademlia::get_providers`. It also fixes (AKA fixes in most cases) the previously prototyped display of Kademlia content keys, in this case `Cid`s.

cc https://github.com/rs-ipfs/rust-ipfs/issues/10

### DIFF
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -95,7 +95,11 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<KademliaEvent> for Behaviour
                             info!("kad: peer {} is close", peer);
                         }
                     }
-                    GetProviders(Ok(GetProvidersOk { key, providers, closest_peers })) => {
+                    GetProviders(Ok(GetProvidersOk {
+                        key,
+                        providers,
+                        closest_peers,
+                    })) => {
                         let key = multibase::encode(Base::Base32Lower, key);
                         if providers.is_empty() && closest_peers.is_empty() {
                             warn!("kad: could not find a provider for {}", key);
@@ -257,9 +261,7 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<BitswapEvent> for Behaviour<
             BitswapEvent::ReceivedWant(peer_id, cid, priority) => {
                 info!(
                     "Peer {} wants block {} with priority {}",
-                    peer_id,
-                    cid,
-                    priority
+                    peer_id, cid, priority
                 );
 
                 let queued_blocks = self.bitswap().queued_blocks.clone();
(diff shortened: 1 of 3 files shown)


tests/kademlia.rs
```diff
@@ -1,51 +1,99 @@
...
+    // connect it to one of the well-known bootstrappers
+    assert!(peer
+        .add_peer(bootstrapper_id, bootstrapper_addr)
         .await
         .is_ok());
 
-    // call kad::bootstrap
-    assert!(extra_peer.bootstrap().await.is_ok());
+    // the Cid of the docs.ipfs.io website
+    let cid: Cid = "bafybeicfjz7woevc5dxvsskibxpxpofkrdjyslbggvvr3d66ddqu744nne"
```
koivunej, line 92: When this breaks, we should fix to the logo svg, which might be a bit more stable: bafkreicncneocapbypwwe3gl47bzvr3pkpxmmobzn7zr2iaz67df4kjeiq

ljedrz: I actually did have it set to the svg at a point, must have accidentally reverted that bit (I was doing some tinkering around it)

ljedrz: done :+1:

src/p2p/behaviour.rs
```diff
@@ -419,16 +420,29 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
+        let key = cid.to_bytes();
+        self.kademlia.get_providers(key.into());
         self.bitswap.want_block(cid, 1);
     }
 
+    // FIXME: it would probably be best if this could return a SubscriptionFuture, so
+    // that the put_block operation truly finishes only when the block is already being
+    // provided; it is, however, pretty tricky in terms of internal communication between
+    // Ipfs and IpfsFuture objects - it would currently require some extra back-and-forth
     pub fn provide_block(&mut self, cid: Cid) {
```
koivunej, line 435: I am not 100% with you on this FIXME but no need to change it now. To elaborate:

"Not 100% with you" as in is that the saving blocks should be as decoupled as possible and providing should only happen for some subset of blocks: possibly only the pinned roots, or pinned roots and their subroots or anything more complex would need to be provided for. Where "subroots" would be the roots of the files in a pinned larger directory structure for the invidiual files.

Decoupling providing and put_block because providing can be quite expensive, and we might not want to do them concurrently at all. Additionally not sure if it's a good idea to plan to use start_providing unless we can come up with a really nice backend for the kademlia store. Otherwise we could just use a simpler and handle the timing ourselves with the "push" api.

koivunej:
Flaky #134 with the readme-doctest again.

src/p2p/behaviour.rs
ljedrz: Fair enough :+1:

PR #261, opened 2020-07-23
update remaining deps

ljedrz:
In order to tackle the test timeout issue I decided to update the remaining dependencies to make sure the root cause is not already fixed upstream.

The only direct dependency that couldn't be updated is `unsigned-varint`, because `cid` depends on the version we are currently using.


ljedrz:
A linker error in Windows? I don't think I have't seen those before, I'll retry when all the others are complete.

ljedrz:
A linker error again; it's in `ipfs-http` and points to `openssl`.

ljedrz:
No more `Cargo.lock` changes to `openssl`, but the Windows build still fails at linking `openssl` :shrug:; package caching?

koivunej:
mac failure is #134.

koivunej:
The windows caching is likely the culprit here. Today I learned vcpkg is in fact using submodules for all of the `ports/` so I am now thinking if we could somehow pin that version. Then using that "pin" as a cache key would be the proper course.

koivunej:
This should be rebased on top of the latest master following #267, which resolved the caching issue.

ljedrz:
The mac failure is the fimeout; if Windows passes, this is good to go.

ljedrz:
Yep, all good - it's good to go @koivunej.

Issue #248, opened 2020-07-16
tests: async-ify mostly async tests


ljedrz:
Some of the tests, e.g. the ones in `connect_two`, are `#[test]` and contain `async_std::task::block_on`, while they could just be `#[async_std::test]`.

aphelionz:
Thanks a lot, @c410-f3r !

PR #262, opened 2020-07-24
Fix the test timeout

ljedrz:
Or rather a handful of drive-bys accumulated while trying to fix it thus far.

While the root cause still eludes me, I've come to a point where `exchange_block` is reliably failing, improving further test conditions. At this point I'm preeetty sure it's not the `SubscriptionRegistry` (which got a fix - the keys for complete `Subscription`s were leaking); for instance, during my fix attempts I had changed the custom `Subscription::Cancelled` value to an `Abortable<SubscriptionFuture<T>>` setup and changed the `Arc` in the `SubscriptionFuture` to a `Weak`, both to no avail.

I'm also confident that it is not fixed with any dependency updates; I was initially testing on top of https://github.com/rs-ipfs/rust-ipfs/pull/261.

My next attempt is probably going to be changing the `Arc` related to the `Ipfs` object to a `Weak` in the `IpfsFuture`, which is now my primary suspect - I think something goes wrong while it is being dropped.

cc https://github.com/rs-ipfs/rust-ipfs/issues/248 (during some fix attempts the `connect_two` tests were failing, hence https://github.com/rs-ipfs/rust-ipfs/commit/8569c9a7038504f5bcd70a770e76e3025e006ddf)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -628,6 +628,13 @@ impl<Types: IpfsTypes> Ipfs<Types> {
     }
 }
 
+impl<T: IpfsTypes> Drop for Ipfs<T> {
+    fn drop(&mut self) {
+        self.repo.shutdown();
+        let _ = self.to_task.clone().try_send(IpfsEvent::Exit);
+    }
+}
+
 /// Background task of `Ipfs` created when calling `UninitializedIpfs::start`.
 // The receivers are Fuse'd so that we don't have to manage state on them being exhausted.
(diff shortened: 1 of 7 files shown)


ljedrz:
Closing this one, since
- test improvements were already done in https://github.com/rs-ipfs/rust-ipfs/pull/265
- @koivunej has the `Subscription` leak fix in the upcoming branch

PR #263, opened 2020-07-27
move bitswap Stats directly under Bitswap

ljedrz:
This makes the bitswap `Stats` persistent between peer disconnects.

In addition, remove the unused and no longer compatible `Ledger` tests.


bitswap/src/behaviour.rs
```diff
@@ -168,6 +224,7 @@ impl NetworkBehaviour for Bitswap {
     fn inject_connected(&mut self, peer_id: &PeerId) {
         debug!("bitswap: inject_connected {}", peer_id);
         let ledger = Ledger::new();
+        self.stats.insert(peer_id.clone(), Default::default());
```
koivunej, line 227: This should be removed in the `inject_disconnected`. However I am just wondering ... Why do we have separate Stats instead of a single `Arc<Stats>` object? I guess the [`/bitswap/ledger?arg=peerid`](https://docs.ipfs.io/reference/http/api/#api-v0-bitswap-ledger) hints at peer level accounting.. Aah, now I remember, it's about the sessioning of the peers when we want to download multiple blocks.

With this overall change, the work that needs to be done in `inject_disconnected` is now to merge the per-ledger Stats into a "global one" but that again gets a bit complicated with the `Arc<Stats>` and unsureness of when it'll be dropped if ever.. Phew, this is a bit complicated again.

ljedrz: Good catch, that line should be removed :+1:. Can you make it into a suggestion so I can delete it on-the-fly?

koivunej: That'd be outside the changed hunks so no I cannot create a comment there. I don't think `self.stats.remove(peer_id)` would be sufficient implementation as then the stats would go down when peers disconnect? Those should only increase.

ljedrz: Ah yeah, right; as for the other point, how about the following, persistence-friendly solution?
```
     pub fn connect(&mut self, peer_id: PeerId) {
-        self.stats.insert(peer_id.clone(), Default::default());
+        self.stats.entry(peer_id.clone()).or_default();
         if self.target_peers.insert(peer_id.clone()) {
             self.events.push_back(NetworkBehaviourAction::DialPeer {
                 peer_id,
(1/2) Stage this hunk [y,n,q,a,d,j,J,g,/,e,?]? y
@@ -224,7 +224,7 @@ impl NetworkBehaviour for Bitswap {
     fn inject_connected(&mut self, peer_id: &PeerId) {
         debug!("bitswap: inject_connected {}", peer_id);
         let ledger = Ledger::new();
-        self.stats.insert(peer_id.clone(), Default::default());
+        self.stats.entry(peer_id.clone()).or_default();
         self.connected_peers.insert(peer_id.clone(), ledger);
         self.send_want_list(peer_id.clone());
```

koivunej: Yeah I guess the Entry::or_default() fixes at least the monotonous stats issue; didn't really think about it happening with the `insert`. So we'll leave the disconnect fixes for a later PR.

Just a note which crossed my mind; I am not so sure when is the `connect` needed, at all. Why should it insert/entry+or_default to stats? If the peer *is not* connected, there will be inject_connected upon succesful connecting, so adding it to `connect` seems redundant. We do nothing with the stats while *connecting*.

ljedrz: If that function is always hit upon connection, then there is no need for the other one indeed. I'll push an update shortly.

ljedrz: Done :+1:.

ljedrz:
`repo::fs::tests::test_fs_blockstore` failed; don't think I've seen that one before, but seems unrelated.

koivunej:
Yeah rare windows filesystem issues are normal :) I've seen that happen once or twice.

koivunej:
Hmm it's the access denied again! Interesting.

bors retry

PR #264, opened 2020-07-27
move the Repo from Ipfs to IpfsFuture

ljedrz:
As far as I can tell from the comments, this has been the plan / target design for a while now. These changes carry a few important improvements:
- `Ipfs` is no longer generic :tada: (a huge advantage for IPFS-as-a-library)
- no more back-and-forth between `Ipfs` and `IpfsFuture` via `RepoEvent`s
- the `Ipfs` methods become uniform in the way they work
- there is no longer a need for `Ipfs` to be within an `Arc` (though it might be preferable for future-proofing)

This is still a work in progress, as: 
- [x] 2 `Ipfs` methods remain to be converted
- [x] 2 `RepoEvent` functionalities need to be re-introduced
- [x] the crate tests need to be adjusted
- [x] the `wantlist_and_cancellation` test needs to be fixed

The conformance tests (excluding the aforementioned missing bits) agree with these changes.

### DIFF
--- a/src/subscription.rs
+++ b/src/subscription.rs
@@ -4,12 +4,11 @@
 //! sharing the same unique numeric identifier, the `SubscriptionId`.
 
 use async_std::future::Future;
-use async_std::task::{self, Context, Poll, Waker};
+use async_std::task::{Context, Poll, Waker};
 use core::fmt::Debug;
 use core::hash::Hash;
 use core::pin::Pin;
 use futures::channel::mpsc::Sender;
-use futures::lock::Mutex;
 use libipld::Cid;
 use libp2p::{kad::QueryId, Multiaddr};
 use std::collections::HashMap;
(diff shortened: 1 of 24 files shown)


src/subscription.rs
```diff
@@ -19,13 +17,27 @@ use std::fmt;
...
 static GLOBAL_REQ_COUNT: AtomicU64 = AtomicU64::new(0);
 
+impl TryFrom<RequestKind> for IpfsEvent {
+    type Error = &'static str;
+
+    fn try_from(req: RequestKind) -> Result<Self, Self::Error> {
+        if let RequestKind::GetBlock(cid) = req {
+            Ok(IpfsEvent::CancelBlock(cid))
+        } else {
+            Err("logic error: RepoEvent can only be created from a Request::GetBlock")
```
ljedrz, line 36:
```suggestion
            Err("logic error: IpfsEvent can only be created from a Request::GetBlock")
```

ljedrz:
Closing this one for now, as it's a spicy change set; we might go back to this approach in the future.

PR #265, opened 2020-07-27
Use async macro for tests

### LINKED ISSUE #248: tests: async-ify mostly async tests

c410-f3r:
Fixes #248

### DIFF
--- a/http/Cargo.toml
+++ b/http/Cargo.toml
@@ -10,6 +10,7 @@ prost-build = "0.6.1"
 vergen = "3.1.0"
 
 [dependencies]
+async-std = { default-features = false, features = ["attributes"], version = "1.6" }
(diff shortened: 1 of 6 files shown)


http/Cargo.toml
```diff
@@ -10,6 +10,7 @@ prost-build = "0.6.1"
 vergen = "3.1.0"
 
 [dependencies]
+async-std = { default-features = false, features = ["attributes"], version = "1.6" }
```
c410-f3r, line 13: No additional here dependency because `async-std` is a transient dependency of `ipfs`

Issue #200, opened 2020-06-23
ipfs-unixfs: walker API has unnecessary unwraps


koivunej:
When walking on a nested directory, the `ContinuedWalk::Directory` can appear multiple times for the same path. It is a bit tricky to see how it's communicated whether or not this is the first time we are walking on this path. The "knowledge" is currently encoded such that `Entry::metadata` will be `Some` on the root of the directory (first time the path is seen) and `None` for successive times. For example, see [ipfs-http/src/v0/root_files.rs](https://github.com/rs-ipfs/rust-ipfs/blob/a1614011a330d32842352bf1095219e6b068b92a/http/src/v0/root_files.rs#L163-L172).

This needs to be encoded in a better way. Perhaps in a way which would allow better matching overall:

```
visit = match walker.continue_walk(&data, &mut cache)? {
    ContinuedWalk::File(segment, file_item) => {
        let total_size = file_item.total_size(); // no unwrap
        let metadata = file_item.metadata(); // no unwrap
        ...
    }
    ContinuedWalk::Directory(DirectoryItem::Root(item)) => {
        let metadata = item.metadata(); // no unwrap
        ...
    }
    ContinuedWalk::Directory(DirectoryItem::Continuation(item)) => item.into_inner(),
    ContinuedWalk::Symlink(bytes, item) => {
        ...
    }
};
```

The 0.0.1 version would allow similar unwrap-less processing though matching `item.as_entry()` but with `enum ContinuedWalk {}` it requires double matching.

Best case scenario would be to find a way which requires only a single match to access either the Entry variant values without unwrapping in ipfs-unixfs or user code. The reason why the initial version did not have a better version of this was that it seemed all options would be self-referential structs or similar, but I am confident there's at least a better way.

koivunej:
Once the final PR related to this lands, I think we are good releasing the 0.1.0 unless objected, added the `needs changelog` tag so that I might *remember* this. Also much appreciated if @c410-f3r would add a one liner for this ergonomics work when we feel like this is done. I try to get the `add` for plain directories hammered out in the meantime.

PR #272, opened 2020-07-30
Split Entry into two different enums to remove unnecessary expects

c410-f3r:
Second take on #200 

`Entry` is now composed of `Bucket(...)` and `Metadata(MetadataEntry { ... })` to separate things that have and doesn't have metadata, which avoids returning optional values derived from a single entry-point.

Next PR will address the newly introduced double matching while iterating over `continue_walk` but it will require deeper logical changes.

### DIFF
--- a/unixfs/src/walk.rs
+++ b/unixfs/src/walk.rs
@@ -10,6 +10,43 @@ use std::convert::TryFrom;
 use std::fmt;
 use std::path::{Path, PathBuf};
 
+/// Representation of the current item of Walker or the last observed item with medatada.
+#[derive(Debug)]
+pub enum MetadataEntry<'a> {
+    /// Current item is a non-root plain Directory or a HAMTShard directory.
+    Directory(&'a Cid, &'a Path, &'a Metadata),
+    /// Current item is possibly a root file with a path, metadata, and a total file size.
+    File(&'a Cid, &'a Path, &'a Metadata, u64),
+    /// Current item is the root directory (HAMTShard or plain Directory).
+    RootDirectory(&'a Cid, &'a Path, &'a Metadata),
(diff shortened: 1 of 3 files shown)


koivunej:
hello @c410-f3r and thanks for your continued work with #200. this is looking like a good step forward. could you separate the `walk.rs` reordering into a separate commit or keep the old ordering to make the reviewing a bit easier?

unixfs/src/walk.rs
```diff
@@ -483,16 +490,23 @@ impl InnerEntry {
...
+            InnerKind::Symlink(cid) => {
+                Entry::Metadata(MetadataEntry::Symlink(cid, &self.path, &self.metadata))
             }
-            RootBucket(cid) => Entry::Directory(cid, &self.path, &self.metadata),
-            Bucket(cid) => Entry::Bucket(cid, &self.path),
-            Directory(cid) => Entry::Directory(cid, &self.path, &self.metadata),
-            File(cid, _, sz) => Entry::File(cid, &self.path, &self.metadata, *sz),
-            Symlink(cid) => Entry::Symlink(cid, &self.path, &self.metadata),
         }
     }
```
koivunej, line 511: In hindsight this seems "really obvious" change, great idea!

Thinking of this naming or meaningwise the difference between the cases might be more on the `Entry::Metadata` meaning "is standalone entry [which might have non-empty metadata]" and `Entry::Bucket` meaning "is continuation [or 'just more links']" but as I think the doc comments on `Entry` variants already cover this detail nicely so no need to change this *right now*.

c410-f3r:
@koivunej Thanks

That is all regarding removing `expect("")`. My current local changes are now solely related to avoid double matching but it will take some time to complete.

PR #274, opened 2020-07-31
fix bogus empty bitswap messages

ljedrz:
We've been seeing `(empty message)` bitswap for a while, but these logs were bogus - this PR fixes those, and adds a small bitswap stress test that can be used for future bitswap debugging.

Also removes some noisy logging and, as a drive-by, fixes a subscription-related log.

### DIFF
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -237,6 +237,8 @@ impl NetworkBehaviour for Bitswap {
     fn inject_event(&mut self, source: PeerId, _connection: ConnectionId, message: MessageWrapper) {
         let mut message = match message {
             // we just sent an outgoing bitswap message, nothing to do here
+            // FIXME: we could commit any pending stats accounting for this peer now
+            // that the message may have sent, if we'd do such accounting
             MessageWrapper::Tx => return,
             // we've received a bitswap message, process it
             MessageWrapper::Rx(msg) => msg,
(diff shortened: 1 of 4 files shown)


bitswap/src/behaviour.rs
```diff
@@ -234,7 +234,14 @@ impl NetworkBehaviour for Bitswap {
         //self.connected_peers.remove(peer_id);
     }
 
-    fn inject_event(&mut self, source: PeerId, _connection: ConnectionId, mut message: Message) {
+    fn inject_event(&mut self, source: PeerId, _connection: ConnectionId, message: MessageWrapper) {
+        let mut message = match message {
+            // we just sent an outgoing bitswap message, nothing to do here
```
koivunej, line 239:
```suggestion
            // we just sent an outgoing bitswap message, nothing to do here
            // FIXME: we could commit any pending stats accounting for this peer now
            // that the message may have sent, if we'd do such accounting
```

tests/bitswap_stress_test.rs
```diff
@@ -0,0 +1,48 @@
+use ipfs::{Block, Node};
+use libipld::cid::{Cid, Codec};
+use multihash::Sha2_256;
+
+fn filter(i: usize) -> bool {
+    i % 2 == 0
+}
+
+#[async_std::test]
+async fn bitswap_stress_test() {
```
koivunej, line 10: could you add a comment or something here, to describe how does this in fact stress the bitswap, maybe what kind of tracing are you expected to see and so on?

ljedrz: ah yeah, good point; it can be `#[ignore]`d in general, it's for manual debugging

PR #276, opened 2020-08-01
Remove unused dependencies

c410-f3r:
First commit only adds `default-features = false` where applicable while the second commit future-proofs possible upstream modifications for every single crate.

```
* bitswap

cargo build: 199 -> 179
cargo check: 212 -> 191
cargo test: 201 -> 181


* ipfs

cargo build: 419 -> 341
cargo check: 433 -> 354
cargo test: 445 -> 359


* ipfs-http

cargo build: 518 -> 408
cargo check: 556 -> 429
cargo test: 522 -> 412
```

By looking into the differences, it is possible to see astonishing results across all operations.


ljedrz:
Sweet, this reduces the size of our `Cargo.lock` by nearly 25% :tada:.

I'm definitely ok with this, just going to give @koivunej a chance to have a look at it too since it's a pretty big change, albeit an internal/meta one.

koivunej:
Sad that our ci build times have so much noise in them that we cannot use them to measure.

So do all of the crates have features or how did you determine where to put `default_features = false`?

Impressive changes on the `http` crates `cargo tree`!

Issue #105, opened 2020-03-22
Connecting to multiaddrs with peerids


koivunej:
After #104 there are multiaddr cases which either do not return an error or hang the http request at least:

1. `GET /swarm/connect?arg=/ip4/127.0.0.1/tcp/4001/ipfs/peerid` is invalid query string (as expected)
2. `GET /swarm/connect?arg=/ip4/127.0.0.1/tcp/4001/p2p/peerid` hangs, doesn't connect according to logging
3. `GET /swarm/connect?arg=/ip4/127.0.0.1/tcp/4001/p2p/typoed_peerid` hangs, doesn't connect according to logging

Where `/ip4/127.0.0.1/tcp/4001` is a local `go-ipfs-0.4.23` and `peerid` is it's peer id, `typoed_peerid` the same but with with single different character.

I didn't investigate these further nor is it clear we need this functionality for the conformance tests. 

Could be that this work can be factored out to parts:

1. if rust-libp2p already gives an error for addresses with p2p in them we are hanging instead of catching that case, which we should be doing either way (the second test case above)
2. dialing an expected peerid with peerid separated from the multiaddr
3. backwards compatible peerid extraction from `/ipfs/peerid` at the `ipfs-http` level

I've understood that rust-libp2p implements dialing to a expected peerid by getting a `NetworkBehaviourAction::DialPeer { peer_id }` followed by offering of addresses for this peerid at `NetworkBehaviour::addresses_of_peer`.

Related discussion over at rust-libp2p: https://github.com/libp2p/rust-libp2p/issues/1462

As the `/ipfs/peerid` seems to be deprecated in favor of `/p2p/peerid` per [comment](https://github.com/libp2p/rust-libp2p/issues/471#issuecomment-420712108) it'd make sense to handle this deprecation (`s!/ipfs/!/p2p/!` style handling) only at `ipfs-http` to get a parsing multiaddr and then move on separating the peerid at some point before asking the Swarm to dial that peerid. There will be some complications and/or duplicate work in `ipfs::p2p::swarm::SwarmApi` related to these two paths. If those cannot be easily dealt with, we could go with requiring the peerid in the multiaddr as ipfs does.

koivunej:
Even with #108 the second case still hangs. This is probably the case of dropped result here: https://github.com/libp2p/rust-libp2p/blob/v0.16.2/swarm/src/lib.rs#L449

aphelionz:
This is 1.2 or later, yes?

[...]

PR #277, opened 2020-08-03
Add the means to connect by PeerId

ljedrz:
I've encountered this while working on improved bitswap <> kademlia interop - we still can't easily connect using `PeerId`. Improve this by introducing an `enum ConnectionTarget` that is either a `Multiaddr` or a `PeerId` and adjusting the `SwarmApi.connect_registry` accordingly. There's also a new related test.

As a drive-by, adjust some `tracing` features that are among the defaults but we'd still enjoy having :smiley:.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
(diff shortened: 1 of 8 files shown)


koivunej:
Does this fix https://github.com/rs-ipfs/rust-ipfs/issues/105, should I merge #110 afterwards? Can't remember if there was something else.

ljedrz:
As far as I can tell this PR is a prerequisite for those 2; while it allows us to connect to known `PeerId`s, it doesn't yet recognize any prefixes etc. I'll take a look at these a bit later, thanks for referring me to them.

ljedrz:
Ok, since the extension to this PR is giving me a little bit of an issue, let's merge this one as-is so as not to block anything else.

bors r+

PR #278, opened 2020-08-03
Update filetime

c410-f3r:
Sets `0.2.12` as the minimum required version because of https://github.com/alexcrichton/filetime/issues/62


ljedrz:
Thanks for noticing that issue!

ljedrz:
A timeout in `kademlia_popular_content_discovery`; it happens from time to time.

bors retry

ljedrz:
Ah, I missed the fact that bors hasn't given up yet :smiley:.

PR #279, opened 2020-08-04
Allow connecting by peer id + addresses

ljedrz:
An extension of https://github.com/rs-ipfs/rust-ipfs/pull/277. I'm not 100% sure how to proceed here, so bear with me.

The main assumption here is that connecting by `PeerId` + `Multiaddr`(s) is the preferable means of connecting to IPFS nodes, though doing that via `libp2p` makes me wonder about the approach - the reason being that `NetworkBehaviourAction`s dial options are **either** by _a_ `Multiaddr` or by a `PeerId`.

The choice I've made here was to attempt multiple connections by `Multiaddr` if the `PeerId`+`Vec<Multiaddr>` combo is provided. The fancy `PartialEq` and `Hash` implementations are needed in order to satisfy the needs of the `SubscriptionRegistry` applicable to connections; the current assumption is that any successful connection that is characterized by the given `PeerId` or any of the `Multiaddr`s (sent with it) fulfills the subscription to that connection request.

### DIFF
--- a/src/p2p/swarm.rs
+++ b/src/p2p/swarm.rs
@@ -6,7 +6,6 @@ use libp2p::swarm::protocols_handler::{
 };
 use libp2p::swarm::{self, DialPeerCondition, NetworkBehaviour, PollParameters, Swarm};
 use std::collections::{HashMap, HashSet, VecDeque};
-use std::hash::{Hash, Hasher};
 use std::time::Duration;
 
 /// A description of currently active connection.
@@ -20,43 +19,11 @@ pub struct Connection {
     pub rtt: Option<Duration>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum ConnectionTarget {
     Addr(Multiaddr),
     PeerId(PeerId),
     PeerWithAddrs(PeerId, Vec<Multiaddr>),
-}
-
-impl PartialEq for ConnectionTarget {
-    fn eq(&self, other: &Self) -> bool {
-        match self {
-            Self::Addr(addr) => match other {
-                Self::PeerWithAddrs(_, addrs) => addrs.contains(addr),
-                Self::Addr(peer_addr) => peer_addr == addr,
-                Self::PeerId(_) => false,
-            },
-            Self::PeerId(id) => match other {
-                Self::PeerWithAddrs(peer_id, _) => peer_id == id,
-                Self::PeerId(peer_id) => peer_id == id,
-                Self::Addr(_) => false,
-            },
-            Self::PeerWithAddrs(id, addrs) => match other {
-                Self::PeerId(peer_id) => peer_id == id,
-                Self::Addr(peer_addr) => addrs.contains(peer_addr),
-                Self::PeerWithAddrs(peer_id, _) => peer_id == id,
-            },
-        }
-    }
-}
-impl Eq for ConnectionTarget {}
-
-impl Hash for ConnectionTarget {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        match self {
-            Self::Addr(addr) => addr.hash(state),
-            Self::PeerId(id) | Self::PeerWithAddrs(id, _) => id.hash(state),
-        }
-    }
 }
 
 impl From<Multiaddr> for ConnectionTarget {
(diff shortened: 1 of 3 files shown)


src/p2p/swarm.rs
```diff
@@ -108,15 +148,26 @@ impl SwarmApi {
...
+                self.events.push_back(NetworkBehaviourAction::DialPeer {
+                    peer_id: id.clone(),
+                    condition: DialPeerCondition::Disconnected,
+                })
+            }
+            ConnectionTarget::PeerWithAddrs(ref _peer_id, ref addrs) => {
+                for addr in addrs {
+                    self.events.push_back(NetworkBehaviourAction::DialAddress {
+                        address: addr.clone(),
+                    })
```
koivunej, line 167: I wonder if the addresses could just be stored for the PeerId in some swarmapi variable and these should always yield just DialPeer, with DialPeerCondition::Disconnected?

ljedrz: Hmm, but doesn't this assume that dialing the `PeerId` is preferable? This puts plenty of trust in the DHT; maybe I could do that and put a timeout on that attempt, so that the addresses can be used as a fallback method?

koivunej: I think dialing the PeerId is always preferable.

> This puts plenty of trust in the DHT

I was not suggesting changing kademlia state; SwarmAPI can provide the addresses through it's NetworkBehaviour implementation: https://github.com/rs-ipfs/rust-ipfs/blob/b099deb4aa9d0dcee9d140d0c1f203941ff4fd37/src/p2p/swarm.rs#L212-L218.

> maybe I could do that and put a timeout on that attempt, so that the addresses can be used as a fallback method?

Just to clarify DHT should be used when user uses `/api/v0/swarm/connect?arg=/p2p/peer_id` but we don't support that at the moment at all. Whenever a req like `/api/v0/swarm/connect?arg=/.../p2p/peer_id` comes in, we should use the provided addresses. This was done in a bit of haste back then so it might not fit well into the current implementation.

I however do not know how does this work with *kademlia* knowing a address for this peer. Should kademlia addresses always be attempted last, so that users provided addresses are preferred? Quite likely. Without looking at the NetworkBehaviour derive impl, I would guess that this ordering is "specified" by ordering kademlia as the last field.

ljedrz: > DHT should be used when user uses /api/v0/swarm/connect?arg=/p2p/peer_id

Ok, all clear now; I just thought this might have been used in the `/api/v0/swarm/connect?arg=/.../p2p/peer_id` cases instead.

ljedrz:
I updated the PR based on the comments; one thing left is potentially ignoring aforementioned further dial attempts to the same target if one is already taking place, but this check would probably be more suited for the `SubscriptionRegistry` - though then we probably shouldn't support multiple identical `Subscription`s (as we currently do in order to satisfy some conformance tests), so I dunno :stuck_out_tongue:.

ljedrz:
Fixed the test that had been working with the earlier iteration of this change set (when the `PeerId`+`Multiaddr`s combo was dialing by the addresses) by reverting it to its original form.

src/p2p/swarm.rs
```diff
@@ -112,10 +119,12 @@ impl SwarmApi {
...
-            ConnectionTarget::PeerId(ref id) => NetworkBehaviourAction::DialPeer {
-                peer_id: id.clone(),
-                condition: DialPeerCondition::Disconnected,
-            },
+            ConnectionTarget::PeerId(ref id) | ConnectionTarget::PeerWithAddrs(ref id, _) => {
+                NetworkBehaviourAction::DialPeer {
+                    peer_id: id.clone(),
+                    condition: DialPeerCondition::Disconnected,
+                }
+            }
```
koivunej, line 127: Just thought about this: what if we are already connected to the PeerId? I think this impl will hang indefinitely until there's a disconnect. I think the way to go would be an "ready subscription", perhaps?

koivunej: Perhaps this could be outside of this PR.

koivunej:
Perhaps rebase/get rid of the conflict (I can't really guess what that conflict is)? Or add a test for that hang and #[ignore] it if its not easy to fix right away?

ljedrz:
I guess I can test it in another PR - this has already deviated far enough from its initial state :stuck_out_tongue:.

PR #280, opened 2020-08-04
Patch a Bitswap leak

ljedrz:
We currently don't properly clean up after a connection to a `Bitswap` peer is closed, which leads to leaking `Ledger`s; funnily enough, it seems that just uncommenting a pre-existing line solves the issue. In addition, add a relevant test.

### DIFF
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -86,7 +86,7 @@ impl Stats {
 pub struct Bitswap {
     /// Queue of events to report to the user.
     events: VecDeque<NetworkBehaviourAction<Message, BitswapEvent>>,
-    /// List of peers to send messages to.
+    /// List of prospect peers to connect to.
     target_peers: FnvHashSet<PeerId>,
     /// Ledger
     pub connected_peers: HashMap<PeerId, Ledger>,
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -231,7 +231,7 @@ impl NetworkBehaviour for Bitswap {
 
     fn inject_disconnected(&mut self, peer_id: &PeerId) {
         debug!("bitswap: inject_disconnected {:?}", peer_id);
-        //self.connected_peers.remove(peer_id);
+        self.connected_peers.remove(peer_id);
     }
 
     fn inject_event(&mut self, source: PeerId, _connection: ConnectionId, message: MessageWrapper) {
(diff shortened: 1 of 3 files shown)


bitswap/src/behaviour.rs
```diff
@@ -231,7 +231,7 @@ impl NetworkBehaviour for Bitswap {
 
     fn inject_disconnected(&mut self, peer_id: &PeerId) {
         debug!("bitswap: inject_disconnected {:?}", peer_id);
-        //self.connected_peers.remove(peer_id);
+        self.connected_peers.remove(peer_id);
```
koivunej, line 234: Could add a comment explaining why stats is not removed as well? I can't guess it from the context.

bitswap/src/behaviour.rs
```diff
@@ -86,7 +86,7 @@ impl Stats {
 pub struct Bitswap {
     /// Queue of events to report to the user.
     events: VecDeque<NetworkBehaviourAction<Message, BitswapEvent>>,
-    /// List of peers to send messages to.
+    /// List of prospect peers to connect to.
```
koivunej, line 89: Does this mean the peers which bitswap has dialed for or what exactly? I am pretty sure it's not the "send messages to" nor has it been for a long time :)

ljedrz: Yeah, I'm wondering if this object shouldn't be removed entirely, as it is only applied in connection attempts from `Bitswap`, and not used for anything else as far as I can tell.

bitswap/src/behaviour.rs
ljedrz: > I wonder if there are previous block reads which were requested but should now be cancelled for this peerid?

Which ones do you have in mind? The related `Ledger` is dropped, and the `get_block`-related futures are peer-agnostic.

koivunej: Oki, because of this line I was thinking that the block reads are in fact per peer: https://github.com/rs-ipfs/rust-ipfs/blob/master/src/p2p/behaviour.rs#L272

Great if those are shared already, but then I would expect there to be some kind of "ticketing" system, perhaps a `HashMap<Cid, Vec<PeerId>>`  kind of thing which would allow you to correlate back the loaded block in case it's wanted by multiple peers.

bitswap/src/behaviour.rs
ljedrz: Ok, it does makes some sense, as it limits the number of `DialPeer` attempts. Still needs to be removed, but that's material for a follow-up.

bitswap/src/behaviour.rs
ljedrz: As discussed, let's defer that for now.

ljedrz:
rebased; bors r+

ljedrz:
bors retry

PR #281, opened 2020-08-04
Some Kademlia debugging

ljedrz:
These changes were sparked by the following 2 observations:
- the logs sometimes indicate that a Kademlia query was executed twice
- `finish_subscription` didn't always result in futures being awoken

While the former remains a mystery (its occurrence is not correlated with subscriptions, meaning it's either some polling issue that eludes me or a bug in `libp2p`), investigating the `kad`<>`Subscription` route yielded a few improvements, namely:
- tweaked `kad` log levels
- proper `SubscriptionFuture` handling in `put_block`
- improved `SubscriptionRegistry` logs and a small fix
- simpler `SubscriptionFuture` type handling (it always returned a `Result`, so now it's a default)
- a `debug_assert` checking that we don't trigger zero-wake cases in `finish_subscription` during tests
- improved `swarm_api` test (that sometimes caused issues with the new `debug_assert`, but could be improved regardless)

### DIFF
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -433,13 +433,19 @@ impl<Types: IpfsTypes> Behaviour<Types> {
         &mut self,
         cid: Cid,
     ) -> Result<SubscriptionFuture<(), String>, anyhow::Error> {
-        let key = cid.to_bytes();
-        match self.kademlia.start_providing(key.into()) {
-            // Kademlia queries are marked with QueryIds, which are most fitting to
-            // be used as kad Subscription keys - they are small and require no
-            // conversion for the applicable finish_subscription calls
-            Ok(id) => Ok(self.kad_subscriptions.create_subscription(id.into(), None)),
-            Err(e) => Err(anyhow!("kad: can't provide block {}: {:?}", cid, e)),
+        // currently disabled; see https://github.com/rs-ipfs/rust-ipfs/pull/281#discussion_r465583345
+        // for details regarding the concerns about enabling this functionality as-is
+        if false {
+            let key = cid.to_bytes();
+            match self.kademlia.start_providing(key.into()) {
+                // Kademlia queries are marked with QueryIds, which are most fitting to
+                // be used as kad Subscription keys - they are small and require no
+                // conversion for the applicable finish_subscription calls
+                Ok(id) => Ok(self.kad_subscriptions.create_subscription(id.into(), None)),
+                Err(e) => Err(anyhow!("kad: can't provide block {}: {:?}", cid, e)),
+            }
+        } else {
+            Err(anyhow!("providing blocks is currently unsupported"))
         }
     }
 
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -427,20 +427,14 @@ impl<Types: IpfsTypes> Behaviour<Types> {
         self.bitswap.want_block(cid, 1);
     }
 
-    // FIXME: it would probably be best if this could return a SubscriptionFuture, so
-    // that the put_block operation truly finishes only when the block is already being
-    // provided; it is, however, pretty tricky in terms of internal communication between
-    // Ipfs and IpfsFuture objects - it would currently require some extra back-and-forth
-    pub fn provide_block(&mut self, cid: Cid) {
+    pub fn provide_block(
+        &mut self,
+        cid: Cid,
+    ) -> Result<SubscriptionFuture<Result<(), String>>, anyhow::Error> {
         let key = cid.to_bytes();
         match self.kademlia.start_providing(key.into()) {
-            Ok(_id) => {
-                // Ok(self.kad_subscriptions.create_subscription(id.into(), None))
-            }
-            Err(e) => {
-                error!("kad: can't provide block {}: {:?}", cid, e);
-                // Err(anyhow!("kad: can't provide block {}", key))
-            }
+            Ok(id) => Ok(self.kad_subscriptions.create_subscription(id.into(), None)),
+            Err(e) => Err(anyhow!("kad: can't provide block {}: {:?}", cid, e)),
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/subscription.rs
```diff
@@ -393,22 +405,20 @@ impl<TRes: Debug + PartialEq> Drop for SubscriptionFuture<TRes> {
...
         }
     }
 }
 
-impl<TRes: Debug + PartialEq> fmt::Debug for SubscriptionFuture<TRes> {
+impl<T: Debug + PartialEq, E: Debug> fmt::Debug for SubscriptionFuture<T, E> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         write!(
             fmt,
             "SubscriptionFuture<Output = Result<{}, Cancelled>>",
```
ljedrz, line 420: TODO: fix the `Cancelled` bit

src/p2p/swarm.rs
```diff
@@ -212,8 +208,10 @@ impl NetworkBehaviour for SwarmApi {
         }
         self.connections.remove(closed_addr);
         // FIXME: should be an error
```
ljedrz, line 210: TODO: remove the `FIXME`

src/p2p/behaviour.rs
```diff
@@ -84,15 +84,15 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<KademliaEvent> for Behaviour
                     }
                     GetClosestPeers(Ok(GetClosestPeersOk { key: _, peers })) => {
                         for peer in peers {
-                            info!("kad: peer {} is close", peer);
+                            debug!("kad: peer {} is close", peer);
```
koivunej, line 87: I remember thinking this earlier but not commenting; why is the key not printed out here? If it is because some other log message always prints it out, perhaps it could be commented here? Same question applies for other places which don't report the key, like the `GetClosestPeers(Err(...))` below.

src/p2p/behaviour.rs
```diff
@@ -427,20 +427,14 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
-    // that the put_block operation truly finishes only when the block is already being
-    // provided; it is, however, pretty tricky in terms of internal communication between
-    // Ipfs and IpfsFuture objects - it would currently require some extra back-and-forth
-    pub fn provide_block(&mut self, cid: Cid) {
+    pub fn provide_block(
+        &mut self,
+        cid: Cid,
+    ) -> Result<SubscriptionFuture<(), String>, anyhow::Error> {
         let key = cid.to_bytes();
         match self.kademlia.start_providing(key.into()) {
```
koivunej, line 435: I did not previously notice that start_providing was already being called. I don't think we should be doing that at all at the moment given the "client dht" vs. "server dht" completion.

Or how do you see the risk of putting a provider record in without having peers who know how to connect to us (and also knowing that we can be connected to)? I think this was the reason why go-ipfs 0.5 introduced the big DHT changes.

To reiterate, I am worried we might be poisoning the dht *unless* we are sure we can be connected to and are making sure our peerid => multiaddr[] record is in place.

src/p2p/behaviour.rs
```diff
@@ -427,20 +427,14 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
-            Ok(_id) => {
-                // Ok(self.kad_subscriptions.create_subscription(id.into(), None))
-            }
-            Err(e) => {
-                error!("kad: can't provide block {}: {:?}", cid, e);
-                // Err(anyhow!("kad: can't provide block {}", key))
-            }
+            Ok(id) => Ok(self.kad_subscriptions.create_subscription(id.into(), None)),
```
koivunej, line 436: So this id is not at all related to the provided cid? Could add a comment why that's not needed or done at the moment?

ljedrz: I'll comment on this :+1:

src/p2p/behaviour.rs
ljedrz: In this case we're just asking for peers closest to us, so the key is just our node's id - I just didn't think it's that interesting in that case.

src/p2p/behaviour.rs
ljedrz: Will those records not be readily available when the other peers connect to us? Kademlia should be re-publishing them periodically. Other than that I don't have any strong feelings towards this, I can comment it out so that it can be reintroduced at a later stage.

src/p2p/behaviour.rs
koivunej: Aah yes ... Didn't remember the implicit "query is for our peerid". Perhaps it could commented here what the expectation is?

src/p2p/behaviour.rs
koivunej: I don't know but I assume the publishing or the record putting is an act of finding all of the closest peers to key, and asking them to store it until expiration. Continuing on what I expect, we should only be providing these records in the event of us being one of the K closest peers to the key, but I guess there could be an optimization.

Probably best to do something like `if false { ... }` around this, with a comment. Is the key correct? I would had assumed that multihash was already enonugh? Perhaps we've only tried it v0 cids which have some compatible byte format?

ljedrz: Fair enough, I've disabled that functionality for now.

As for the `Cid` working as a key, it all depends on how the other implementations are doing it I guess; if this is pure `Multihash` then of course we should change it; I'll do it in a follow-up if that's the case, as the `want_block` part will need to be adjusted too.

koivunej: This looks good.

src/p2p/behaviour.rs
koivunej: So what was the outcome here?

ljedrz: I commented on it:
```
                // Kademlia queries are marked with QueryIds, which are most fitting to
                // be used as kad Subscription keys - they are small and require no
                // conversion for the applicable finish_subscription calls
```

PR #282, opened 2020-08-04
Convert unixfs to no_std

### LINKED ISSUE #247: ipfs-unixfs: no_std and alloc -ification for wasm use
Creating this issue following interest expressed in the comments of #222. I don't have much to go on from my [previous comment on this topic](https://github.com/rs-ipfs/rust-ipfs/issues/222#issuecomment-656238585).

c410-f3r:
Fixes #247

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ cid = { default-features = false, version = "0.5" }
 dirs = { default-features = false, version = "3.0" }
 domain = { default-features = false, features = ["resolv"], git = "https://github.com/nlnetlabs/domain", rev="084964" }
 futures = { default-features = false, features = ["compat", "io-compat"], version = "0.3.5" }
-ipfs-unixfs = { path = "unixfs" }
+ipfs-unixfs = { features = ["std", "with-filetime"], path = "unixfs" }
 libipld = { branch = "update_cid", default-features = false, features = ["dag-pb", "dag-json"], git = "https://github.com/ljedrz/rust-ipld" }
 libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-async-std", "mdns", "mplex", "ping", "secio", "yamux"], version = "0.22" }
 multibase = { default-features = false, version = "0.8" }
--- a/unixfs/Cargo.toml
+++ b/unixfs/Cargo.toml
@@ -9,7 +9,9 @@ repository = "https://github.com/rs-ipfs/rust-ipfs"
 version = "0.0.1"
 
 [features]
-default = ["filetime"]
+default = []
+std = []
+with-filetime = ["filetime"]
 
 [dependencies]
 cid = { default-features = false, version = "0.5" }
--- a/unixfs/src/walk.rs
+++ b/unixfs/src/walk.rs
@@ -1,26 +1,32 @@
-use crate::dir::{ShardError, UnexpectedDirectoryProperties};
-use crate::file::visit::{Cache, FileVisit, IdleFileVisit};
-use crate::file::{FileError, FileReadFailed};
-use crate::pb::{FlatUnixFs, PBLink, ParsingFailed, UnixFsType};
-use crate::{InvalidCidInLink, Metadata, UnexpectedNodeType};
+use crate::{
+    dir::{ShardError, UnexpectedDirectoryProperties},
+    file::{
+        visit::{Cache, FileVisit, IdleFileVisit},
+        FileError, FileReadFailed,
+    },
+    pb::{FlatUnixFs, PBLink, ParsingFailed, UnixFsType},
+    InvalidCidInLink, Metadata, UnexpectedNodeType,
+};
+use alloc::{
+    borrow::{Cow, ToOwned},
+    string::String,
+    vec::Vec,
+};
 use cid::Cid;
+use core::{convert::TryFrom, fmt};
 use either::Either;
-use std::borrow::Cow;
-use std::convert::TryFrom;
-use std::fmt;
-use std::path::{Path, PathBuf};
 
 /// Representation of the current item of Walker or the last observed item with medatada.
 #[derive(Debug)]
 pub enum MetadataEntry<'a> {
     /// Current item is a non-root plain Directory or a HAMTShard directory.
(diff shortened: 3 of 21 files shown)


unixfs/src/walk.rs
```diff
@@ -511,20 +517,16 @@ impl InnerEntry {
     }
 
     fn set_path(&mut self, name: &str, depth: usize) {
-        debug_assert_eq!(self.depth, self.path.ancestors().count());
-
         while self.depth >= depth && self.depth > 0 {
-            assert!(self.path.pop());
+            assert!(path_pop(&mut self.path));
```
c410-f3r, line 521: This is a poor man's PathBuf::pop. If not good enough, it is possible to port std's PathBuf::(ancestors|pop) to a internal module for private usage.

unixfs/src/walk.rs
```diff
@@ -847,8 +851,22 @@ impl fmt::Display for Error {
...
 
+fn path_pop(path: &mut String) -> bool {
+    if path.is_empty() || path == "/" {
+        return false;
+    }
+    let truncate_idx = if let Some(last_slash_idx) = path.trim_end_matches('/').rfind('/') {
+        last_slash_idx
+    } else {
+        0
+    };
```
koivunej, line 865: I just rewrote the same full path growing on a string to build the directory trees (draft PR later today), mine ended up using `full_path.bytes().rposition(|ch| ch == b'/')`. Can you see when could that `trim_end_matches('/')` be needed in general or in this file?

Could you maybe add a direct test case for this? Can't remember a case when the path would had been just `/` however I've been looking at that (draft PR later today) so can't really remember any details here.

Agreed with [your comment](https://github.com/rs-ipfs/rust-ipfs/pull/282/files#r465363929) on the place of use that we might need to get that `poor_mans_string_paths` internal library going if no one has done it before. At least w.r.t. the unixfs paths (or dag-pb links) are also serialized as utf8 so ... Well, I guess there's the case of go (libraries?) happily serializing non-utf8 bytes as utf8. Might have to see someone report a bug on such dag-pb and then decide what to do for it. Until then utf8 paths are probably more than enough.

Now thinking this side aloud the "path" handling is always "unix" when it comes to separators. Though, separators are only present in the (draft PR later today) input and only here: the Walker API aims to provide **platform compatible tracking of paths** (so that the files can be exported). I wonder if this no_std-ification should take this to account, or should we have some intermediatery path representation which can be always transformed to the *platform paths*? I can't think of anything else than windows being the problematic one, as usual :)

unixfs/Cargo.toml
```diff
@@ -9,7 +9,9 @@ repository = "https://github.com/rs-ipfs/rust-ipfs"
 version = "0.0.1"
 
 [features]
-default = ["filetime"]
+default = []
+std = []
+with-filetime = ["filetime"]
```
koivunej, line 14: is creating a `with-` prefixed feature for a dependency a better practice than just using the deps as features? I can't remember seeing this style.

Cargo.toml
```diff
@@ -21,7 +21,7 @@ cid = { default-features = false, version = "0.5" }
 dirs = { default-features = false, version = "3.0" }
 domain = { default-features = false, features = ["resolv"], git = "https://github.com/nlnetlabs/domain", rev="084964" }
 futures = { default-features = false, features = ["compat", "io-compat"], version = "0.3.5" }
-ipfs-unixfs = { path = "unixfs" }
+ipfs-unixfs = { features = ["std", "with-filetime"], path = "unixfs" }
```
koivunej, line 24: Leaving `with-filetime` or `filetime` out would allow you to shave at least one optional dependency off; I don't think this is used from anywhere.

unixfs/src/walk.rs
koivunej: Started a new thread on the path_pop. In general I agree with the `poor_mans_string_paths` library now that I had to again write this myself. Before porting lets make sure this makes sense (in the other thread).

unixfs/Cargo.toml
c410-f3r: This is a generic pattern that avoids conflicts with the dependency name and is useful for features that have more than one dependency.

For example, `with-serde = ['serde', 'serde_derive']` (works) instead of `serde = ['serde', 'serde_derive']` (error).

unixfs/src/walk.rs
c410-f3r: > I just rewrote the same full path growing on a string to build the directory trees (draft PR later today), mine ended up using full_path.bytes().rposition(|ch| ch == b'/'). Can you see when could that trim_end_matches('/') be needed in general or in this file?

https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.components, specifically `A trailing slash is normalized away, /a/b and /a/b/ are equivalent`.

> Now thinking this side aloud the "path" handling is always "unix" when it comes to separators. Though, separators are only present in the (draft PR later today) input and only here: the Walker API aims to provide platform compatible tracking of paths (so that the files can be exported). I wonder if this no_std-ification should take this to account, or should we have some intermediatery path representation which can be always transformed to the platform paths? I can't think of anything else than windows being the problematic one, as usual :)

The algorithm was made taking into account `utf8` and `unix` paths only. If Windows interoperability is a hard-constraint, then a standalone and multi-platform "no_std" PathBuf must first be created and it will probably take some time.

c410-f3r:
Closing since this matter can't be solved without Windows support

ljedrz:
How about instead of creating an `std` feature adding a `no_std`-enabled one that applies everywhere besides Windows? I doubt Windows users are the ones targeted by the `no_std` feature :smiley:.

unixfs/Cargo.toml
koivunej: Thanks for clarifying that up, sounds like the way to go.

aphelionz:
@c410-f3r What do you think of the above suggestion? I think this will still be viable even if it didn't work on Windows

c410-f3r:
Rust features are meant to be additive, therefore, a `windows` feature would be more appropriated but it isn't too different from the `std` feature in the current situation. Furthermore, the recent additions on master makes `no_std` compatibility more challenging although not impossible.

PR #285, opened 2020-08-05
Make IpldDag and Ipns async_traits

ljedrz:
This was a part of https://github.com/rs-ipfs/rust-ipfs/pull/264 which was surprisingly painless and seems more idiomatic than the current approach of creating ad-hoc intermediate `struct`s containing an `Ipfs` (which is internally an `Arc`).


ljedrz:
Oof, forgot the tests :smile:.

PR #287, opened 2020-08-06
Enable 2 conformance timeout tests

ljedrz:
I noticed that we have 2 seemingly simple conformance tests disabled. I wanted to take a stab at fixing them, but try as I may, I can't trigger any hangs, so they might already have been fixed by some of our recent changes; let's see how they fare with the CI.

### DIFF
--- a/conformance/test/index.js
+++ b/conformance/test/index.js
@@ -35,9 +35,6 @@ const factory = createFactory(options)
 tests.miscellaneous(factory, { skip: [
   'dns',
   'resolve',
-  // these cause a hang 20% of time:
-  'should respect timeout option when getting the node id',
-  'should respect timeout option when getting the node version',
   // this hangs on windows, see #251
   'stop'
 ] })


ljedrz:
A `kad` timeout and some `npm` 404s :roll_eyes:; will rerun once all builds complete.

ljedrz:
bors retry

koivunej:
What has changed for these tests to no longer hang?

ljedrz:
Good question - no clue, but I can't reproduce the hangs at all :thinking:.

koivunej:
I think last time they happened on windows and I guess you are not testing on windows, which prompted the addition of why-is-node-running and later the creation of https://github.com/ipfs/js-ipfsd-ctl/issues/525 and the following up PR (which we have patched at the moment). Note regarding the timeout tests, and why I don't think they are too important at the moment: https://github.com/ipfs/js-ipfs/issues/3161.

Lets do a few windows runs.

ljedrz:
> I think last time they happened on windows

Yeah, I'm on Linux so I wouldn't have this issue; if it still happens, I'll add this info to the comment, as I don't think this was specified (otherwise I wouldn't have attempted this at all :stuck_out_tongue:).

ljedrz:
Aaah, this still doesn't have that kad fix :laughing:; I'll rebase.

koivunej:
Re-running.

ljedrz:
Re-running again.

ljedrz:
Closing this as we're not convinced that passing these tests properly displays progress.

PR #290, opened 2020-08-06
Some maintenance

ljedrz:
Some maintenance work so as not to conflict too much with the pending PRs:
- depend on cid instead of libipld where only Cid is used
- update the `domain` dep (big `Cargo.lock` wins)

cc #75

### DIFF
--- a/src/ipns/dns.rs
+++ b/src/ipns/dns.rs
@@ -1,14 +1,14 @@
 use crate::error::Error;
 use crate::path::IpfsPath;
-use domain::core::iana::Rtype;
-use domain::core::rdata::Txt;
-use domain::core::{Dname, Question};
-use domain::resolv::stub::resolver::Query;
-use domain::resolv::{Resolver, StubResolver};
-use futures::compat::{Compat01As03, Future01CompatExt};
+use bytes::Bytes;
+use domain::base::iana::Rtype;
+use domain::base::{Dname, Question};
+use domain::rdata::rfc1035::Txt;
+use domain_resolv::{stub::Answer, StubResolver};
 use futures::future::{select_ok, SelectOk};
 use futures::pin_mut;
 use std::future::Future;
+use std::io;
 use std::pin::Pin;
 use std::str::FromStr;
 use std::task::{Context, Poll};
@@ -18,8 +18,10 @@ use thiserror::Error;
 #[error("no dnslink entry")]
 pub struct DnsLinkError;
 
+type FutureAnswer = Pin<Box<dyn Future<Output = Result<Answer, io::Error>>>>;
+
 pub struct DnsLinkFuture {
-    query: SelectOk<Compat01As03<Query>>,
+    query: SelectOk<FutureAnswer>,
 }
 
 impl Future for DnsLinkFuture {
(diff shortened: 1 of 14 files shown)


src/ipns/dns.rs
```diff
@@ -57,16 +59,18 @@ impl Future for DnsLinkFuture {
...
+    let qname = Dname::<Bytes>::from_str(&dnslink)?;
     let question = Question::new_in(qname, Rtype::Txt);
-    let query2 = StubResolver::new().query(question).compat();
+    let resolver = StubResolver::new();
+    let query2 = Box::pin(async move { resolver.query(question).await });
 
     Ok(DnsLinkFuture {
-        query: select_ok(vec![query1, query2]),
+        query: select_ok(vec![query1 as FutureAnswer, query2]),
     }
```
koivunej, line 74: did you try either of the test_resolve1 or test_resolve2?

tests/wantlist_and_cancellation.rs
```diff
@@ -2,10 +2,10 @@ use async_std::{
     future::{pending, timeout},
     task,
 };
+use cid::Cid;
 use futures::future::{select, Either, FutureExt};
 use futures::future::{AbortHandle, Abortable};
 use ipfs::Node;
-use libipld::Cid;
```
koivunej, line 8: Great idea; I had been thinking that these were already removed but apparently not!

src/ipns/dns.rs
koivunej: In test_resolve2 we should just expect an IpfsPath and that's it. I did not know of the IpfsPath when writing that FIXME :)

tests/wantlist_and_cancellation.rs
ljedrz: I already did a pass before, but left some stragglers apparently :stuck_out_tongue:

src/ipns/dns.rs
ljedrz: just tried it:
```
test ipns::dns::tests::test_resolve1 ... ok
test ipns::dns::tests::test_resolve2 ... FAILED

failures:

---- ipns::dns::tests::test_resolve2 stdout ----
thread 'ipns::dns::tests::test_resolve2' panicked at 'assertion failed: `(left == right)`
  left: `"/ipfs/bafybeidzfuvdzwkvfv4eonmbtpvkxkwxfxdfowww6wijxoj2chhryf7sc4"`,
 right: `"/ipfs/QmbV3st6TDZVocp4H2f4KE3tvLP1BEpeRHhZyFL9gD4Ut4"`', src/ipns/dns.rs:93:9


failures:
    ipns::dns::tests::test_resolve2
```

ljedrz: So I guess resolving works now :stuck_out_tongue:? I can update the path, but should I enable the tests? Probably not, as they could share the same fate as that one `kad` test we disabled today, no?

koivunej: Great that those work now as well! I this whole thing could be simplified a lot, but probably best to throw this in as is right away. No need to enable that test.. Need to learn if we can have tests enabled only when !ci.

PR #293, opened 2020-08-07
set opt-level = 1 in the dev profile

ljedrz:
This should allow https://github.com/rs-ipfs/rust-ipfs/pull/284 to pass the conformance tests without `ipfs-http` triggering a stack overflow.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,9 @@ license = "MIT OR Apache-2.0"
 name = "ipfs"
 version = "0.1.0"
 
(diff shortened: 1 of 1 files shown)


koivunej:
I think we should not do this because of the long build times. At least while the fixes in a3f27e5 are enough :)

ljedrz:
Agreed - since https://github.com/rs-ipfs/rust-ipfs/pull/284 no longer triggers stack overflows, a workaround is not needed.

PR #294, opened 2020-08-09
Remove Walker indirection

### LINKED ISSUE #200: ipfs-unixfs: walker API has unnecessary unwraps

c410-f3r:
Fixes #200.

By using mutable references instead of taking ownership, the following is now possible:

```rust
ContinuedWalk::Symlink(bytes, _, path, metadata) => {
  ...
}

// Instead of

ContinuedWalk::Symlink(bytes, item) => {
  if let Entry::Metadata(metadata_entry) = item.as_entry() {
    ...
  }
}
```


unixfs/examples/get.rs
```diff
@@ -70,49 +70,44 @@ fn walk(blocks: ShardedBlockStore, start: &Cid) -> Result<(), Error> {
...
-                    let (seconds, _) = metadata.mtime().unwrap_or((0, 0));
-                    println!(
-                        "s {:o} {:>12} {:>16} {:?} -> {:?}",
-                        mode, seconds, "-", path, target
-                    );
-                }
-                item.into_inner()
+            // presense of metadata can be used to determine if this is the first apperiance of
+            // a directory by looking at the metadata: sibling hamt shard buckets do not have
+            // metadata.
```
koivunej, line 91: since this comment is still present ... how exactly are non-root buckets differentiated from directories?

koivunej: aah, it would be `ContinuedWalk::Directory(Entry::Bucket(_))`, I'll add a suggestion below

unixfs/examples/get.rs
```diff
@@ -70,49 +70,44 @@ fn walk(blocks: ShardedBlockStore, start: &Cid) -> Result<(), Error> {
...
+            ContinuedWalk::Directory(Entry::Metadata(metadata_entry)) => {
+                let metadata = metadata_entry.metadata();
+                let path = metadata_entry.path();
+                let mode = metadata.mode().unwrap_or(0o0755) & 0o7777;
+                let (seconds, _) = metadata.mtime().unwrap_or((0, 0));
+                println!("d {:o} {:>12} {:>16} {:?}", mode, seconds, "-", path);
+            }
+            ContinuedWalk::Symlink(bytes, Entry::Metadata(metadata_entry)) => {
```
koivunej, line 99:
```suggestion
            ContinuedWalk::Directory(bytes, Entry::Bucket(..)) => {
                // continuation of a HAMT sharded directory. should usually just be ignored.
            }
            ContinuedWalk::Symlink(bytes, Entry::Metadata(metadata_entry)) => {
```

unixfs/examples/get.rs
```diff
@@ -70,49 +70,44 @@ fn walk(blocks: ShardedBlockStore, start: &Cid) -> Result<(), Error> {
...
-                    let target = Path::new(std::str::from_utf8(bytes).unwrap());
-                    let mode = metadata.mode().unwrap_or(0o0755) & 0o7777;
-                    let (seconds, _) = metadata.mtime().unwrap_or((0, 0));
-                    println!(
-                        "s {:o} {:>12} {:>16} {:?} -> {:?}",
-                        mode, seconds, "-", path, target
-                    );
-                }
-                item.into_inner()
+            // presense of metadata can be used to determine if this is the first apperiance of
```
koivunej, line 89:
```suggestion
```

unixfs/examples/get.rs
```diff
@@ -70,49 +70,44 @@ fn walk(blocks: ShardedBlockStore, start: &Cid) -> Result<(), Error> {
...
+            // a directory by looking at the metadata: sibling hamt shard buckets do not have
```
koivunej, line 90:
```suggestion
```

unixfs/examples/get.rs
koivunej, line 91: oh the horror of lack of multiline suggestions :)
```suggestion
```

unixfs/examples/get.rs
```diff
@@ -70,49 +70,44 @@ fn walk(blocks: ShardedBlockStore, start: &Cid) -> Result<(), Error> {
...
+                let path = metadata_entry.path();
+                let target = Path::new(std::str::from_utf8(bytes).unwrap());
+                let mode = metadata.mode().unwrap_or(0o0755) & 0o7777;
+                let (seconds, _) = metadata.mtime().unwrap_or((0, 0));
+                println!(
+                    "s {:o} {:>12} {:>16} {:?} -> {:?}",
+                    mode, seconds, "-", path, target
+                );
             }
+            _ => {}
```
koivunej, line 110: I think with my suggestions above this would now be an error.
```suggestion
```

koivunej: If this suggestion is followed it leaves the `File(_, Bucket(_, _))` unhandled. I don't think this variant should be possible, due to how the "continuations" carry over the metadata for files but not for buckets.

c410-f3r: I see now. Thanks for clarifying it

koivunej: Oki, looking at this now, resolving other comment branches. This can be resolved, I'll try continue a proper review or a PR against your code.

c410-f3r:
Based on all comments and what is currently on master:

```rust
pub fn as_entry(&self) -> Entry<'_> {
    match &self.kind {
        InnerKind::Bucket(cid) => Entry::Bucket(cid, &self.path),
        InnerKind::Directory(cid) => {
            Entry::Metadata(MetadataEntry::Directory(cid, &self.path, &self.metadata))
        }
        InnerKind::File(cid, _, sz) => {
            Entry::Metadata(MetadataEntry::File(cid, &self.path, &self.metadata, *sz))
        }
        InnerKind::RootBucket(cid) => {
            Entry::Metadata(MetadataEntry::Directory(cid, &self.path, &self.metadata))
        }
        InnerKind::RootDirectory(cid) | InnerKind::BucketAtRoot(cid) => Entry::Metadata(
            MetadataEntry::RootDirectory(cid, &self.path, &self.metadata),
        ),
        InnerKind::Symlink(cid) => {
            Entry::Metadata(MetadataEntry::Symlink(cid, &self.path, &self.metadata))
        }
    }
}
```

The new `ContinuedWalk` API is now declared as follows:

```rust
pub enum ContinuedWalk<'a> {
    Bucket(&'a Cid, &'a Path),
    Directory(&'a Cid, &'a Path, &'a Metadata),
    File(FileSegment<'a>, &'a Cid, &'a Path, &'a Metadata, u64),
    RootDirectory(&'a Cid, &'a Path, &'a Metadata),
    Symlink(&'a [u8], &'a Cid, &'a Path, &'a Metadata),
}
```

`ContinuedWalk::Bucket` is the "non-continued"/"non-root" HAMT shard while `ContinuedWalk::RootDirectory` agglutinates `InnerKind::RootDirectory`  and `InnerKind::RootBucket` although they could all be `ContinuedWalk::Directory` since its has the same signature.

There are no more "unwraps" or indirect variants but I must say that after this whole process was quite challenging

c410-f3r:
Even with the structural and logical changes, I did my best the keep the original order in the last commit.

unixfs/src/walk.rs
```diff
@@ -10,43 +10,183 @@ use std::convert::TryFrom;
...
+            | Self::File(_, _, p, ..)
+            | Self::RootDirectory(_, p, ..)
+            | Self::Symlink(_, _, p, ..) => p,
+        }
+    }
+}
+
+/// Errors which can occur while walking a tree.
+#[derive(Debug)]
+pub enum Error {
```
koivunej, line 45: Moving `Error` and `ContinuedWalk`, `FileSegment` back to where it was would certainly drop the lines changed.

unixfs/src/walk.rs
```diff
@@ -639,216 +642,73 @@ impl InnerEntry {
...
-            InvalidCid(l) => Error::InvalidCid(l),
-        }
-    }
-}
-
-impl From<UnexpectedDirectoryProperties> for Error {
-    fn from(e: UnexpectedDirectoryProperties) -> Self {
-        Error::UnsupportedDirectory(e)
-    }
+// FIXME: could simplify roots to optinal cid variants?
```
koivunej, line 652:
```suggestion
```

I think this was just done by dropping the Cid the InnerEntry.

unixfs/src/walk.rs
```diff
@@ -639,216 +642,73 @@ impl InnerEntry {
...
+    /// Symlink optionally on the root level
+    Symlink,
 }
 
-impl From<ShardError> for Error {
-    fn from(e: ShardError) -> Self {
-        Error::UnsupportedHAMTShard(e)
-    }
+/// Converts a link of specifically a Directory (and not a link of a HAMTShard).
+fn convert_link(
```
koivunej, line 671: Moving these link converters back where they used to be would be a big win for the lines changed.

unixfs/src/walk.rs
```diff
@@ -356,93 +446,21 @@ impl Walker {
 
 /// Represents what the `Walker` is currently looking at. Converted to `Entry` for public API.
 #[derive(Debug)]
-struct InnerEntry {
+pub(crate) struct InnerEntry {
```
koivunej, line 449:
```suggestion
struct InnerEntry {
```

This doesn't seem like a required change.

unixfs/src/walk.rs
```diff
@@ -356,93 +446,21 @@ impl Walker {
 
 /// Represents what the `Walker` is currently looking at. Converted to `Entry` for public API.
```
koivunej, line 447:
```suggestion
/// Represents what the `Walker` is currently looking at.
```

unixfs/src/walk.rs
```diff
@@ -356,93 +446,21 @@ impl Walker {
...
-    Bucket(Cid),
-    /// Directory on any level except root
-    Directory(Cid),
-    /// File optionally on the root level
-    File(Cid, Option<FileVisit>, u64),
-    /// Symlink optionally on the root level
-    Symlink(Cid),
-}
-
-impl fmt::Debug for InnerKind {
```
koivunej, line 390: Why was this not returned for InnerEntry?

unixfs/src/walk.rs
```diff
@@ -546,16 +546,16 @@ impl InnerEntry {
...
-                self.kind = RootBucket(cid);
+            InnerKind::Bucket
+            | InnerKind::BucketAtRoot
+            | InnerKind::Directory
+            | InnerKind::File(None, ..)
+            | InnerKind::RootBucket
+            | InnerKind::RootDirectory
+            | InnerKind::Symlink => {
+                self.cid = cid;
+                self.kind = InnerKind::RootBucket;
```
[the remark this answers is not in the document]
c410-f3r: While trying to understand the code, some time were spend to figure out where these variants were coming from even with the top level import declaration. This declarative direction has a clearer approach and says right away its intention.

It might be silly but for a file that has more than 1000 lines where most of the time you are thinking: What is going on? Every help counts.

Nevertheless, it is just a nit that can be reverted if it is indeed desired.

unixfs/src/walk.rs
c410-f3r: It is usual to keep private elements after public ones and helps with organization and code visibility. For example, many projects like Rustc itself and rust-analyzer also do this.

```
1. fn foo();
2. pub struct A;
3. pub enum B;
4. enum C;
5. pub fn bar();
```
->
```
1. pub struct A;
2. pub enum B;
3. pub fn bar();
4. fn foo();
5. enum C;
```

unixfs/src/walk.rs
c410-f3r: `Cid` was moved to `InnerEntry` to avoid yet another matching when transposing to `ContinuedWalk` and since every variant has `Cid`, it doesn't seemed to be a problem.

In regards to debug representations, this lost formatting can be re-implemented to `InnerEntry` in a following PR.

unixfs/src/walk.rs
ljedrz: That's all fine; however, this kind of change (formatting, ordering) should be separated from functional changes, i.e. contained in its own dedicated commit, so that the functional changes can be reviewed separately from visual ones.

c410-f3r: It is possible to see this ordering solely in the first commit. The other two were order restrained by the maximum of my capacity

unixfs/src/file/visit.rs
```diff
@@ -149,11 +149,11 @@ impl FileVisit {
     ///
     /// Returns on success a tuple of bytes and new version of `FileVisit` to continue the visit,
     /// when there is something more to visit.
-    pub fn continue_walk<'a>(
+    pub fn continue_walk<'a: 'b, 'b>(
         mut self,
         next: &'a [u8],
         cache: &mut Option<Cache>,
-    ) -> Result<(&'a [u8], Option<Self>), FileReadFailed> {
+    ) -> Result<(&'b [u8], Option<Self>), FileReadFailed> {
```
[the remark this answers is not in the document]
c410-f3r: Must have slipped in one of my previous attempts when trying to satisfy the borrow checker

unixfs/src/walk.rs
ljedrz: > It is possible to see this ordering solely in the first commit.
That's true, though due to the fact that it is the first commit and the later ones build on top of it, it is not possible to reverse it to view the whole changeset without it.

In addition, in this case the size of the changes introduced by the reordering is quite big, warranting its very own PR. We have updated our [contributing guidelines](https://github.com/rs-ipfs/rust-ipfs/blob/master/CONTRIBUTING.md) to make it more clear: the patch size should be kept minimal.

unixfs/src/file/visit.rs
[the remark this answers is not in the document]
koivunej: Checked that this is unnecessary. The change you made to the previous `continue_walk` with the three lifetimes is necessary however. In hindsight it seems obvious selection of lifetimes but it must have taken some work to find it :)

unixfs/src/walk.rs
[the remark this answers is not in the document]
koivunej: > While trying to understand the code, some time were spend to figure out where these variants were coming from even with the top level import declaration.

Yeah, I often shuffle code around as well while learning and reading it. I just try to make sure I `git checkout -p` those changes before they end up in my commits; splitting stuff like this out afterwards with `git rebase -i`and `edit` is not impossible but ... It's just easier to do it up front :)

> This declarative direction has a clearer approach and says right away its intention.

I kind of understand this point, and I understand that it's not too often I've seen the enum variants being used however I feel that we are now discussing wheter a specifically added language feature should or should not be used for which it was added for. I don't think this is the forum for such discussions.

koivunej:
@c410-f3r Oki I've now resolved most of the comments I've made and am offering them up as a PR against your PR https://github.com/c410-f3r/rust-ipfs/pull/1.

I hope we can get that PR merged in fast so that we can move on with the review here, then merge this and enjoy the much improved api. While the PR I made is quite difficult to eyeball I hope the changes in diffstats help.

c410-f3r:
Thank you guys for the patience and collaboration. In addition to learning and teaching, arguments and discussions are also great ways that make us think better by incorporating knowledge and experience.

Issue #175, opened 2020-05-25
Default to noise, drop aes-ctr patching


koivunej:
For upcoming go-ipfs 0.6 we could move to noise instead of secio. This would allow dropping aes-ctr patch which may or may not work and then the crate would be releasable.

koivunej:
I think the first thinks to check here are:

 * are the bootstrap nodes going to be updated to support noise or will they be secio/tls only?
 * do the go-ipfs 0.6 and the latest noise interop? 

The local connection test can be done simply with creating a new local go-ipfs instance with `ipfs init`, then starting for example the http into the background, trying to connect the two nodes together (might be easier to connect from go-ipfs).

Not sure how to find out about the bootstrap nodes. If there are still some issues connecting with rust-libp2p to all of the nodes, perhaps go-ipfs could have tls and secio configured off to only try connecting with noise to the nodes?

ljedrz:
Current attempts to substitute `SecioConfig` with `NoiseConfig` in `TcpConfig` result in the following error when tested against a `go-ipfs 0.6` node (`ipfs swarm connect <rust-ipfs node address>`):
```
failed to negotiate security protocol: error unmarshaling remote handshake payload: proto: NoiseHandshakePayload: illegal tag 0 (wire type 0)
```
Relevant section of the logs:
```[2020-06-24T10:11:47Z DEBUG multistream_select::listener_select] Listener: confirming protocol: /noise
[2020-06-24T10:11:47Z DEBUG multistream_select::listener_select] Listener: sent confirmed protocol: /noise
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read state: Init
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read state: ReadLen { buf: [0, 0], off: 0 }
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read state: ReadLen { buf: [0, 0], off: 0 }
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read: next frame len = 32
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read state: ReadData { len: 32, off: 0 }
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read: read 32/32 bytes
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read: decrypting 32 bytes
[2020-06-24T10:11:47Z TRACE libp2p_noise::io] read: payload len = 0 bytes
[...]

PR #300, opened 2020-08-11
use noise authentication with the XX handshake

ljedrz:
The new `0.23` release of `libp2p` seems to have fixed the `noise` incompatibility issues with `go-ipfs`, so update it, drop `SECIO` and use `noise-xx` instead.

While I've successfully tested this both ways against a `go-ipfs-0.6` node, an extra manual test won't hurt :wink:.

fixes https://github.com/rs-ipfs/rust-ipfs/issues/75
fixes https://github.com/rs-ipfs/rust-ipfs/issues/175


koivunej:
manual interop testing: 
 - go-ipfs 0.5 which didn't have noise support doesn't work
 - go-ipfs 0.6.0 can be connected, ping works, life is good

Tried even with the test kademlia_popular_content_discovery, seems to work nicely.

bors r+

PR #305, opened 2020-08-12
Add a subscription stress test

ljedrz:
`#[ignore]`d by default; it currently works pretty well for a reasonable amount of time.


ljedrz:
I should probably add some comments :laughing:.

Issue #275, opened 2020-08-01
ipfs-http crate has two different runtimes


c410-f3r:
Two runtimes: `async-std` from `ipfs` and `tokio` from the crate's internals. This isn't a comfortable situation because of compilation times and additional external dependencies.

If desirable, I am willing to port everything to a single runtime. Just choose the most convenient crate, for example, `tokio` and friends can be replaced with `http-rs` alternatives.

koivunej:
Agreed that the current setup is not ideal, but it has so far been workable. Did you see the hint of [using lld-9](https://github.com/rs-ipfs/rust-ipfs/blob/master/CONTRIBUTING.md#compilation-is-slow-help) in our CONTRIBUTING.md? It of course doesn't save one the compilation time but helps with the huge linker time, at least "on my system".

> If desirable, I am willing to port everything to a single runtime. Just choose the most convenient crate, for example, tokio and friends can be replaced with http-rs alternatives.

`warp` has been so far joy to use so I'd recommend to hold off on any big rewrites for now. Best case scenario would be that the alternatives could use the same "request handler impls" (the non-routing related `async fn`'s, like [`inner_put`](https://github.com/rs-ipfs/rust-ipfs/blob/e224b715ab12bcb942f93aedbe01b1260a2f67bf/http/src/v0/block.rs#L100-L140)). If you already know tide well, could you comment on this?

I've been hoping to have similar alternatives in `ipfs` than the `libp2p` has for using tokio or async-std: https://docs.rs/libp2p/0.22.0/libp2p/tcp/struct.TokioTcpConfig.html, but that of course takes a bit more effort, given the blockstore has to use the same underlying components and we cannot have exclusive features, at least yet. Could be that the implementations just need to be duplicated and use some neat trick to select which one to use if both are enabled. I haven't really looked into this much.

c410-f3r:
@koivunej Thanks for you feedback

More investigation is needed on my side since I am not fully aware about the described nomenclature or the internals machinery. Hope to come up with something feasible in the near future.

PR #307, opened 2020-08-13
Use tokio instead of async_std

ljedrz:
We already use it in `ipfs-http`, our dependencies either already use it or are compatible with it and it has some nice features that we'd like to use.

Note: the `async` tests now specify `max_threads = 1` in order to make sure the multi-threaded executor is not being lenient about any hidden bugs.

fixes https://github.com/rs-ipfs/rust-ipfs/issues/275

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ rand = { default-features = false, features = ["getrandom"], version = "0.7" }
 serde = { default-features = false, features = ["derive"], version = "1.0" }
 serde_json = { default-features = false, features = ["std"], version = "1.0" }
 thiserror = { default-features = false, version = "1.0" }
-tokio = { default-features = false, features = ["full"], version = "0.2" }
+tokio = { default-features = false, features = ["fs", "rt-threaded", "stream"], version = "0.2" }
 tracing = { default-features = false, features = ["log"], version = "0.1" }
 tracing-futures = { default-features = false, features = ["std", "futures-03"], version = "0.2" }
 void = { default-features = false, version = "1.0" }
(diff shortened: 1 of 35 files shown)


Cargo.toml
```diff
@@ -24,14 +23,15 @@ domain = { default-features = false, version = "0.5" }
 domain-resolv = { default-features = false, version = "0.5" }
 futures = { default-features = false, features = ["compat", "io-compat"], version = "0.3.5" }
 ipfs-unixfs = { path = "unixfs" }
-libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-async-std", "mdns", "mplex", "noise", "ping", "yamux"], version = "0.23" }
+libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-tokio", "mdns", "mplex", "noise", "ping", "yamux"], version = "0.23" }
```
koivunej, line 26: TODO: soon tokio-mdns as well, mdns might still use async-std by default.

ljedrz: Good catch! I was looking at the main features in search for other `tokio` ones and found none - `mdns` **does** use `async-std` right now, probably gonna change soon, like you're saying.

ljedrz:
it seems that the `resolve` tests don't like Windows; since this is a **big** PR I'll `#[ignore]` them again for now.

koivunej:
The failure for future reference: https://github.com/rs-ipfs/rust-ipfs/runs/980815860#step:16:64

> it seems that the resolve tests don't like Windows

Did you understand from somewhere else that this could be wider issue? Could be that windows is behind a restricted dns for example or just this particular runner was without global dns.

> Note: the async tests now specify max_threads = 1 in order to make sure the multi-threaded executor is not being lenient about any hidden bugs.

From probably the first bad `tests/` case I've hoped we could go back to tokio for single threaded test cases. Turns out a lot of the test cases have been just flaky/broken but this should stop us from writing any futures which are always ready for example, though it's been a while since I last wrote such. In multi threaded runtime those will be fine, but on single threaded runtime hopefully the runtime should grind to a halt but at minimum it should be noticeable.

koivunej:
I think windows is about to hang again.

bors r+

PR #308, opened 2020-08-13
cargo update; remove unused rand feature

ljedrz:
Just a maintenance update to `Cargo.lock` via `cargo update` so that actual later dependency additions or deletions don't cause unrelated changes to it.


ljedrz:
Also removes the `libpild` fork dependency accidentally left out previously :see_no_evil:.

PR #311, opened 2020-08-13
Remove unneccesary Jekyll file

aphelionz:
This PR removes the _config.yml file, which is not needed since the website content lives in other repos.


aphelionz:
bors retry

PR #314, opened 2020-08-17
Use instrument instead of in_scope in lib

ljedrz:
Apparently `Span::in_scope` is just:
```
pub fn in_scope<F: FnOnce() -> T, T>(&self, f: F) -> T {
    let _enter = self.enter();
    f()
}
```
and `Span::enter` is a no-no in `async` contexts.


ljedrz:
Rebased.

bors retry

PR #315, opened 2020-08-17
Use explicit Multiaddr wrappers

ljedrz:
Whether a `Multiaddr` contains `Protocol::P2p` (i.e. essentially `/p2p/PeerId`) or not has a lot of bearing on conformance tests, the internal object and APIs and error-handling. Since it is easy to introduce related human errors, it would be a good idea to enforce some type safety here by introducing 2 wrapper objects: `MultiaddrWoPeerId` and `MultiaddrWithPeerId`.

The extent to which they should be applied (in lieu of `Multiaddr`) is a matter for discussion - in my initial attempt I decided to apply them quite liberally, as it results in more type safety (there are still a few plain `Multiaddr`s left though).

### DIFF
--- a/http/src/v0/swarm.rs
+++ b/http/src/v0/swarm.rs
@@ -1,5 +1,5 @@
 use super::support::{with_ipfs, StringError};
-use ipfs::{Ipfs, IpfsTypes, Multiaddr};
+use ipfs::{Ipfs, IpfsTypes, MultiaddrWithPeerId};
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
@@ -16,7 +16,7 @@ async fn connect_query<T: IpfsTypes>(
 ) -> Result<impl warp::Reply, warp::Rejection> {
     let target = query
         .arg
-        .parse::<Multiaddr>()
+        .parse::<MultiaddrWithPeerId>()
         .map_err(|e| warp::reject::custom(StringError::from(e)))?;
     ipfs.connect(target)
         .await
@@ -80,8 +80,8 @@ async fn peers_query<T: IpfsTypes>(
                 None
             };
             Peer {
-                addr: conn.address.to_string(),
-                peer: conn.peer_id.to_string(),
+                addr: conn.addr.multiaddr.as_ref().to_string(),
+                peer: conn.addr.peer_id.to_string(),
                 latency,
             }
         })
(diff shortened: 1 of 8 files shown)


src/p2p/swarm.rs
```diff
@@ -1,22 +1,104 @@
...
-use libp2p::swarm::{self, DialPeerCondition, NetworkBehaviour, PollParameters, Swarm};
+use libp2p::swarm::{self, NetworkBehaviour, PollParameters, Swarm};
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::convert::TryFrom;
 use std::time::Duration;
+use std::{fmt, str::FromStr};
+
+/// A wrapper for `Multiaddr` that does **not** contain `Protocol::P2p`.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct MultiaddrWoPeerId(Multiaddr);
```
koivunej, line 18: I think we should not abbreviate here so I'd vote for expanding the "Wo" => "Without". I as a non-native speaker am having doubts using "w/o" ever but capitalized and without the slash it seems quite strange.

http/src/v0/swarm.rs
```diff
@@ -162,14 +162,18 @@ pub fn addrs_local<T: IpfsTypes>(
 
 #[derive(Debug, Deserialize)]
 struct DisconnectQuery {
-    arg: Multiaddr,
+    arg: String,
```
koivunej, line 165: I wonder if the `crate::v0::support::StringSerialized<MultiaddrWithPeerId>` could be used here? It'll probably change the error message to something ackward.

src/p2p/swarm.rs
```diff
@@ -1,22 +1,104 @@
...
+
+impl From<Multiaddr> for MultiaddrWoPeerId {
+    fn from(addr: Multiaddr) -> Self {
+        Self(
+            addr.into_iter()
+                .filter(|p| !matches!(p, Protocol::P2p(_)))
+                .collect(),
```
koivunej, line 25: This will not work for those p2p-circuit multiaddrs correctly at all. It should be a single pop though I dislike this lenient func either way.

ljedrz: Meh, it could also filter out `Protocol::P2pCircuit`; as for liking this or not this is just a matter of opinion, as one _can_ infallibly convert a `Multiaddr` containing `P2p` protocols into one that doesn't (as long as there are other protocols like `Tcp`); making the `Multiaddr` > `MultiaddrWithoutPeerId` fallible makes me wonder about error-handling in the `NetworkBehaviour::inject_*` functions - for now we can probably safely assume they don't contain the `P2p` protocols, since the `PeerId` is passed separately?

http/src/v0/swarm.rs
ljedrz: Yeah, it works :+1:.

ljedrz: ...but the error message is less verbose, so let's leave it as `String` for now, as discussed.

PR #317, opened 2020-08-18
Conformance test suite updates

ljedrz:
`package.json` updates:
```
"interface-ipfs-core": "0.137.0" > "0.139.0"
"ipfs-http-client":    "^44.3.0" > "46.0.0"
"ipfs":                "^0.43.0" > "0.49.0"
```

The new version of the conformance test suite requires 2 new `/version` fields (blank for now) and a patch for one of the `add` tests (as we don't expose `/files` yet). ~~It also allows us to transition the `Repo` to only keep `Multihash`es instead of `Cid`s, as `refs_local` is expected to return `Cid v1` with `Codec::Raw`, which can be produced on the fly.~~ (deferred)


ljedrz:
Forgot about `test_inner_local`; will fix and update shortly

PR #318, opened 2020-08-19
Update libp2p and use TokioMdns

ljedrz:
`libp2p 0.24` enables us to only depend on a single async executor (in our case, `tokio`).

A small change in the code, a big win for `Cargo.lock`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,86 +101,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


ljedrz:
Looks like `aarch64-apple` decided to hang, but since there are no new deps, this should be unrelated.

koivunej (approved):
awesome!

bors r+

PR #319, opened 2020-08-19
Introduce KadResult

ljedrz:
I've been planning to add some of the `/dht` functionalities and since it looks like it's going to be a bit complicated, I'd like to split these upcoming changes into meaningful bits.

First up: the ability to get results from `Kademlia`'s `SubscriptionRegistry`. While working on this I noticed that some of the results can be provided multiple times (e.g. `BootstrapOk` that indicates the number of peers left to bootstrap against), which should solve the mystery of `Kademlia` not adhering to the `debug_assert` we have in `SubscriptionRegistry::finish_subscription`.

As an accidental drive-by there's also a fix to the `Kademlia` bootstrap test which was silently failing after we introduced the `Multiaddr` wrappers.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,7 +49,7 @@ pub use self::error::Error;
 use self::ipns::Ipns;
 pub use self::p2p::pubsub::{PubsubMessage, SubscriptionStream};
 use self::p2p::{create_swarm, SwarmOptions, TSwarm};
-pub use self::p2p::{Connection, MultiaddrWithPeerId};
+pub use self::p2p::{Connection, KadResult, MultiaddrWithPeerId};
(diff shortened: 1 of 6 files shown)


koivunej (approved):
bors r+

koivunej:
I think the test case verification hasn't changed so we could reintroduce the bug here, or did I miss something?

ljedrz:
> I think the test case verification hasn't changed so we could reintroduce the bug here

Yeah, it's possible - I haven't been able to reproduce it at all before, so I couldn't come up with a proper test case; however, this seems like a viable solution, and since that bug happened rarely, I thought we could verify it with the help of CI instead of dedicating too much time to finding a solution to what essentially is a very minor issue.

PR #320, opened 2020-08-19
Implement Ipfs::find_peer and /dht/findpeer

ljedrz:
Builds on https://github.com/rs-ipfs/rust-ipfs/pull/319, only the last 2 commits are new.

Introduces `Ipfs::find_peer` and `/dht/findpeer` that return addresses belonging to the given `PeerId`; if they are connected to already, the values are returned right away, otherwise the local DHT records are checked.

Blocked by https://github.com/rs-ipfs/rust-ipfs/pull/319.

PR #323, opened 2020-08-20
make dht/findpeer search the non-local DHT if necessary

ljedrz:
As promised in https://github.com/rs-ipfs/rust-ipfs/pull/320, this follow-up implements an "active" DHT search as a fallback for the initial local-only mechanism and updates the existing test to properly verify that this works.

A side-effect of the changes is the introduction of `own_id: PeerId` to `IpfsFuture` that we've considered in the past; it might change to the whole `IpfsOptions` in the future, but for now it is enough.


src/lib.rs
```diff
@@ -1011,13 +1037,19 @@ impl<TRepoTypes: RepoTypes> Future for IpfsFuture<TRepoTypes> {
...
+                        let locally_known_addrs = if !swarm_addrs.is_empty() {
+                            swarm_addrs
                         } else {
                             self.swarm.kademlia().addresses_of_peer(&peer_id)
                         };
+                        let addrs = if !locally_known_addrs.is_empty() || local_only {
+                            Either::Left(locally_known_addrs)
+                        } else {
+                            let own_id = self.own_id.clone();
+                            Either::Right(self.swarm.get_closest_peers(own_id))
```
koivunej, line 1051: surprised this is not `peer_id` but `own_id`; could this have some comment explaining why?

ljedrz: not much reason other than me having a brainfart moment, I'm afraid :laughing:; fixed in the force push

PR #331, opened 2020-08-22
Use alloc and core crates where they are applicable

c410-f3r:
First step towards no_std compatibility for unixfs

### DIFF
--- a/src/unixfs/cat.rs
+++ b/src/unixfs/cat.rs
@@ -1,11 +1,11 @@
 use crate::{Error, Ipfs, IpfsTypes};
 use async_stream::stream;
 use cid::Cid;
-use futures::stream::Stream;
-use ipfs_unixfs::file::{visit::IdleFileVisit, FileReadFailed};
 use core::borrow::Borrow;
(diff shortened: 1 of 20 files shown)


koivunej (approved):
This is looking good, apologies for the delay.

bors r+

src/unixfs/cat.rs
```diff
@@ -1,11 +1,11 @@
...
 use async_stream::stream;
 use cid::Cid;
+use core::borrow::Borrow;
+use core::fmt;
+use core::ops::Range;
 use futures::stream::Stream;
 use ipfs_unixfs::file::{visit::IdleFileVisit, FileReadFailed};
-use std::borrow::Borrow;
-use std::fmt;
-use std::ops::Range;
```
koivunej, line 8: Just noticed while handling local conflicts that these were probably extra. `ipfs` should not attempt to be no_std at this time.

c410-f3r: Indeed, the `ipfs` crate deals with input and output operations that are tied to the underlying OS

PR #333, opened 2020-08-23
Remove top-level command item from PR template

aphelionz:
This PR removes the "top-level command" requirement from the PR template. We don't have a CLI yet and we can add it back in when we do.


ljedrz (approved):
LGTM
bors r+

PR #334, opened 2020-08-24
Run (and pass) most of the dht conformance tests

ljedrz:
I think I'll cut the PR at this point already, as I'm seeing some issues related to the skipped tests that could make it grow bigger.

Current status: `12 passing, 8 skipped`.

### DIFF
--- a/conformance/patches/find_provs.patch
+++ b/conformance/patches/find_provs.patch
@@ -1,0 +1,23 @@
+--- node_modules/interface-ipfs-core/src/dht/find-provs.js
++++ node_modules/interface-ipfs-core/src/dht/find-provs.js
+@@ -38,12 +38,13 @@ module.exports = (common, options) => {
+     before('add providers for the same cid', async function () {
+       this.timeout(10 * 1000)
+
+-      const cids = await Promise.all([
+-        nodeB.object.new('unixfs-dir'),
+-        nodeC.object.new('unixfs-dir')
++      const blob = Buffer.from('blorb')
++      const blocks = await Promise.all([
++        nodeB.block.put(blob),
++        nodeC.block.put(blob)
+       ])
+
+-      providedCid = cids[0]
++      providedCid = blocks[0].cid.toString()
+
+       await Promise.all([
+         all(nodeB.dht.provide(providedCid)),
+--
+2.27.0
+
--- a/conformance/patches/find_provs.patch
+++ b/conformance/patches/find_provs.patch
@@ -14,7 +14,7 @@
        ])
 
 -      providedCid = cids[0]
-+      providedCid = blocks[0].cid.toString()
++      providedCid = blocks[0].cid
 
        await Promise.all([
          all(nodeB.dht.provide(providedCid)),
--- a/http/src/v0/dht.rs
+++ b/http/src/v0/dht.rs
@@ -29,7 +29,7 @@ struct ResponsesMember {
 
 #[derive(Debug, Deserialize)]
 pub struct FindPeerQuery {
-    arg: String,
+    arg: StringSerialized<PeerId>,
     // FIXME: doesn't seem to be used at the moment
     verbose: Option<bool>,
     timeout: Option<StringSerialized<humantime::Duration>>,
@@ -44,7 +44,7 @@ async fn find_peer_query<T: IpfsTypes>(
         verbose: _,
         timeout,
     } = query;
-    let peer_id = arg.parse::<PeerId>().map_err(StringError::from)?;
+    let peer_id = arg.into_inner();
     let addrs = ipfs
         .find_peer(peer_id.clone())
         .maybe_timeout(timeout.map(StringSerialized::into_inner))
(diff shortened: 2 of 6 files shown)


conformance/patches/find_provs.patch
```diff
@@ -0,0 +1,23 @@
...
+       this.timeout(10 * 1000)
+
+-      const cids = await Promise.all([
+-        nodeB.object.new('unixfs-dir'),
+-        nodeC.object.new('unixfs-dir')
++      const blob = Buffer.from('blorb')
++      const blocks = await Promise.all([
++        nodeB.block.put(blob),
++        nodeC.block.put(blob)
+       ])
```
koivunej, line 14: I dont think this would be a problematic change for the interface-ipfs-core.

conformance/patches/find_provs.patch
```diff
@@ -0,0 +1,23 @@
...
+
+-      providedCid = cids[0]
++      providedCid = blocks[0].cid.toString()
```
koivunej, line 17: This however might be a problematic change, at least the string conversion part. I think the conversion should be handled by js-ipfs-http-client.

```suggestion
-      providedCid = cids[0]
+      providedCid = blocks[0].cid
```

src/lib.rs
```diff
@@ -1134,12 +1130,12 @@ impl<TRepoTypes: RepoTypes> Future for IpfsFuture<TRepoTypes> {
...
-                    IpfsEvent::GetProviders(key, ret) => {
-                        let future = self.swarm.get_providers(key);
+                    IpfsEvent::GetProviders(cid, ret) => {
+                        let future = self.swarm.get_providers(cid);
                         let _ = ret.send(future);
                     }
-                    IpfsEvent::Provide(key, ret) => {
-                        let _ = ret.send(self.swarm.start_providing(key));
+                    IpfsEvent::Provide(cid, ret) => {
+                        let _ = ret.send(self.swarm.start_providing(cid));
```
koivunej, line 1138: I think we should have the key before providing it (as in: get_block) or error if we dont have it?

At least a comment would be needed here.

koivunej: Oops sorry it was handled above already.

src/p2p/behaviour.rs
```diff
@@ -496,15 +496,17 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
         self.kad_subscriptions
             .create_subscription(self.kademlia.get_providers(key).into(), None)
     }
 
     pub fn start_providing(
         &mut self,
-        key: Key,
+        cid: Cid,
     ) -> Result<SubscriptionFuture<KadResult, String>, anyhow::Error> {
+        let key = Key::from(cid.to_bytes());
```
koivunej, line 509: it amazes me that this conversion is enough but it probably is, for the interface-http-core tests.

Could there be some comment or similar on why this isn't a multihash, or about the future plans of changing it to multihash or something?

conformance/patches/find_provs.patch
ljedrz: didn't think we cared, as I consider this to just be duct-tape to make our tests pass (while preserving the actual thing being tested), but I guess I can change this at `js-ipfs-http-client` level

http/src/v0/dht.rs
```diff
@@ -144,17 +145,16 @@ async fn provide_query<T: IpfsTypes>(
         verbose: _,
         timeout,
     } = query;
-    let key = arg.into_bytes();
-    ipfs.provide(key)
+    let cid = arg.parse::<Cid>().map_err(StringError::from)?;
```
koivunej, line 148: is there an expected error message in the tests which forbids the use of `StringSerialized<Cid>`?

src/p2p/behaviour.rs
ljedrz: yeah, as far as I can remember js-ipfs has not moved on to `Multihash`es yet; I can comment on it

http/src/v0/dht.rs
ljedrz: nah, I just forgot about it; I'll try changing it

src/p2p/behaviour.rs
aphelionz: I believe 0.49 has, or maybe they're planning on it for 0.50...

conformance/patches/find_provs.patch
koivunej: Does the conversion fail without the `.toString()` removed?

ljedrz: it actually doesn't - maybe it was something I experimented with initially and just left it there; regardless, `should be able to find providers` is commented out now so it's not needed in this PR

src/p2p/behaviour.rs
ljedrz: yeah, plain `Multihash` works here too

koivunej (approved):
Looking good!

bors r+

PR #335, opened 2020-08-24
Less panic-ky inject_addr_reach_failure

ljedrz:
It appears that the `unwrap` there was a bit too liberal - try to mitigate it.


koivunej (approved):
Nice, thanks.

bors r+

koivunej:
I cancelled the merge yesterday because I managed to get another panic triggered by some mdns shenanigans and timing on fetch-and-cat example. However I am now unable to get go-ipfs 0.6 to act like it did yesterday. We'll need to come up with some example configs to test this..

koivunej (approved):
Oki, lets go ahead; we'll still need to mdns manual test cases. Maybe I'll run into issues with manual testing..

bors r+

PR #343, opened 2020-08-27
Simplified node creation

ljedrz:
The `UninitializedIpfs` struct neither has to be created using `async` nor is it necessary as a proxy object at all. We can simplify things even further, by not having the proxy `IpfsInner` object either.

This is a draft to "gauge the room" - if this is indeed a preferable course of action, I'll fix the docs, tests etc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -293,7 +293,7 @@ impl<Types: IpfsTypes> UninitializedIpfs<Types> {
     /// The span is attached to all operations called on the later created `Ipfs` along with all
     /// operations done in the background task as well as tasks spawned by the underlying
     /// `libp2p::Swarm`.
-    pub async fn new(options: IpfsOptions, span: Option<Span>) -> Self {
+    pub fn new(options: IpfsOptions, span: Option<Span>) -> Self {
         let repo_options = RepoOptions::from(&options);
         let (repo, repo_events) = create_repo(repo_options);
         let keys = options.keypair.clone();
@@ -308,8 +308,8 @@ impl<Types: IpfsTypes> UninitializedIpfs<Types> {
         }
     }
 
-    pub async fn default() -> Self {
-        Self::new(IpfsOptions::default(), None).await
+    pub fn default() -> Self {
+        Self::new(IpfsOptions::default(), None)
     }
 
     /// Initialize the ipfs node. The returned `Ipfs` value is cloneable, send and sync, and the
@@ -1254,8 +1254,6 @@ mod node {
             let span = Some(Span::current());
 
             let (ipfs, fut): (Ipfs<TestTypes>, _) = UninitializedIpfs::new(opts, span)
-                .in_current_span()
-                .await
                 .start()
                 .in_current_span()
                 .await
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,23 +211,27 @@ impl Default for IpfsOptions {
     }
 }
 
-#[derive(Debug)]
-pub struct Ipfs<Types: IpfsTypes>(Arc<IpfsInner<Types>>);
-
-impl<Types: IpfsTypes> Clone for Ipfs<Types> {
-    fn clone(&self) -> Self {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -277,88 +281,57 @@ enum IpfsEvent {
...
-    pub async fn default() -> Self {
-        Self::new(IpfsOptions::default(), None).await
+    pub fn default() -> Self {
+        Self::new(IpfsOptions::default(), None)
     }
 
     /// Initialize the ipfs node. The returned `Ipfs` value is cloneable, send and sync, and the
     /// future should be spawned on a executor as soon as possible.
-    pub async fn start(self) -> Result<(Ipfs<Types>, impl Future<Output = ()>), Error> {
+    pub async fn start(&mut self) -> Result<impl Future<Output = ()>, Error> {
```
koivunej, line 312: The story behind of `UninitializedIpfs` was probably to originally work around some Send or Sync bounds related to some type which is no longer an issue however I think the real benefit of `UninitalizedIpfs` was that there was no `start` on `Ipfs` which would be called multiple times as the move semantics of the lang give rise to nicer APIs.

What does one do with an `Ipfs` value if the `start` fails?

Could the `start` be moved to IpfsOptions...? As a `build` method or something? Or maybe in place of `new`?

src/lib.rs
```diff
@@ -277,88 +281,57 @@ enum IpfsEvent {
...
-impl<Types: IpfsTypes> UninitializedIpfs<Types> {
+impl<Types: IpfsTypes> Ipfs<Types> {
     /// Configures a new UninitializedIpfs with from the given options and optionally a span.
     /// If the span is not given, it is defaulted to `tracing::trace_span!("ipfs")`.
     ///
     /// The span is attached to all operations called on the later created `Ipfs` along with all
     /// operations done in the background task as well as tasks spawned by the underlying
     /// `libp2p::Swarm`.
-    pub async fn new(options: IpfsOptions, span: Option<Span>) -> Self {
+    pub fn new(options: IpfsOptions, span: Option<Span>) -> Self {
```
koivunej, line 291: Working on the fetch_and_cat the second None looks quite confusing, and I have to look it up (probably because I didn't introduce it but :)). **Since** this line is being changed right now, I wish we could do something for this... Maybe add a method on IpfsOptions? `fn in_span(self, span: Span) -> IpfsOptions`?

src/lib.rs
```diff
@@ -211,25 +211,29 @@ impl Default for IpfsOptions {
...
-        Ipfs(Arc::clone(&self.0))
-    }
-}
-
 /// Ipfs struct creates a new IPFS node and is the main entry point
 /// for interacting with IPFS.
 #[derive(Debug)]
-pub struct IpfsInner<Types: IpfsTypes> {
+pub struct Ipfs<Types: IpfsTypes> {
     pub span: Span,
```
koivunej, line 218: I have missed this earlier, why is this `pub`?

ljedrz: the `Span`? good point; I'll check if this can be removed

src/lib.rs
ljedrz: > What does one do with an Ipfs value if the start fails?

same applies to `UninitializedIpfs`, no? I'm all for improving the naming here, I'll think about it

koivunej: Well `UninitializedIpfs` is a worthless value in itself except for starting, and `start` consumed itself, so you don't have a value left.

ljedrz: while working on ipfs-substrate I had a case where it was easy to spawn tasks, but you had no control over what they were returning, making the setup practical for long-running tasks; in this case the benefit is that `Ipfs` can be created (and kept) in a non-`async` context, while the background task can be concealed away; actually, since `Ipfs` is `Clone`, this can still consume `self`, would that be better?

ljedrz:
Obviated by https://github.com/rs-ipfs/rust-ipfs/pull/345.

PR #344, opened 2020-08-28
Better DHT docs and tests

ljedrz:
Not much to add here - the tests were mostly refactored and some got improved as a side effect.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -725,7 +725,9 @@ impl<Types: IpfsTypes> Ipfs<Types> {
     }
 
     /// Obtain the addresses associated with the given `PeerId`; they are first searched for locally
-    /// and the DHT is used as a fallback.
+    /// and the DHT is used as a fallback: a `Kademlia::get_closest_peers(peer_id)` query is run and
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -782,7 +784,10 @@ impl<Types: IpfsTypes> Ipfs<Types> {
         }
     }
 
-    /// Establishes the local node as a provider of a value for the given Cid.
+    /// Establishes the node as a provider of a block with the given Cid: it publishes a provider
+    /// record with the given key (Cid) and the node's PeerId to the peers closest to the key. The
+    /// publication of provider records is periodically repeated as per the interval specified in
+    /// `libp2p`'s  `KademliaConfig`.
     pub async fn provide(&self, cid: Cid) -> Result<(), Error> {
```
koivunej, line 791: I wonder if this should include a warning on "provides are not persisted across launches". Probably not. I'd like to implement it in the near future.

koivunej: We may need to want to create an issue tracking of all things persistent. Is there /remind functionality on github? :) I'll add a note for myself.

koivunej (approved):
Muchos better.

bors r+

PR #345, opened 2020-08-28
Simplified node internals and creation

ljedrz:
Obviates https://github.com/rs-ipfs/rust-ipfs/pull/343, which was similar, but also removed the `UninitializedIpfs` object.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,23 +211,25 @@ impl Default for IpfsOptions {
     }
 }
 
-#[derive(Debug)]
-pub struct Ipfs<Types: IpfsTypes>(Arc<IpfsInner<Types>>);
-
-impl<Types: IpfsTypes> Clone for Ipfs<Types> {
-    fn clone(&self) -> Self {
-        Ipfs(Arc::clone(&self.0))
-    }
-}
-
 /// Ipfs struct creates a new IPFS node and is the main entry point
 /// for interacting with IPFS.
 #[derive(Debug)]
-pub struct IpfsInner<Types: IpfsTypes> {
+pub struct Ipfs<Types: IpfsTypes> {
     span: Span,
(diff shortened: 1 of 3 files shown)


ljedrz:
It's ready to review, but I'm waiting with adjusting tests until it's determined we want all these changes.

src/lib.rs
```diff
@@ -1251,11 +1251,7 @@ mod node {
         }
 
         pub async fn with_options(opts: IpfsOptions) -> Self {
-            let span = Some(Span::current());
-
-            let (ipfs, fut): (Ipfs<TestTypes>, _) = UninitializedIpfs::new(opts, span)
-                .in_current_span()
-                .await
+            let (ipfs, fut): (Ipfs<TestTypes>, _) = UninitializedIpfs::new(opts)
```
koivunej, line 1254: the only problem I see left here is that the method would need to be generic over `<Types: IpfsTypes>` instead of assigning the `TestTypes`.

src/p2p/mod.rs
```diff
@@ -52,7 +52,7 @@ pub async fn create_swarm<TIpfsTypes: IpfsTypes>(
     // Set up an encrypted TCP transport over the Mplex protocol.
     let transport = transport::build_transport(options.keypair.clone());
 
-    let swarm_span = ipfs.0.span.clone();
+    let swarm_span = ipfs.span.clone();
```
koivunej, line 55: I was thinking we maybe should do a `tracing::trace_span!(parent: &ipfs.span, "swarm")` "pass" over the current tracing usage as I am not sure what does this Clone exactly do, but at least it's no longer public member so I am fine with this.

koivunej:
Couldn't push to this branch so closing in favor of rebased #394.

PR #346, opened 2020-08-29
Test go-ipfs DHT interop

ljedrz:
Builds on https://github.com/rs-ipfs/rust-ipfs/pull/344, only the last 5 commits are new.

Adds a `test_dht_with_go` feature which allows to run the DHT tests with a go-ipfs node in the middle of the chain of rust-ipfs nodes, as long as the `GO_IPFS_PATH` environment variable pointing to the go-ipfs binary is provided.

~~As of now, the DHT tests using a go-ipfs node need to be run one at a time; an appropriate FIXME is included.~~

Also changes the test `Node` object to advertise the `/ipfs/lan/kad/1.0.0` kad protocol that is also used by go-ipfs in tests.

Blocked by https://github.com/rs-ipfs/rust-ipfs/pull/344.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,7 +116,7 @@ impl IpfsOptions {
             keypair: Keypair::generate_ed25519(),
             mdns: Default::default(),
             bootstrap: Default::default(),
-            kad_protocol: Default::default(),
+            kad_protocol: Some("/ipfs/lan/kad/1.0.0".to_owned()),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,6 +116,7 @@ impl IpfsOptions {
             keypair: Keypair::generate_ed25519(),
             mdns: Default::default(),
             bootstrap: Default::default(),
+            // default to lan kad for go-ipfs use in tests
             kad_protocol: Some("/ipfs/lan/kad/1.0.0".to_owned()),
         }
     }
--- a/tests/kademlia.rs
+++ b/tests/kademlia.rs
@@ -129,7 +129,7 @@ fn start_go_node() -> GoIpfsNode {
     // GO_IPFS_PATH should point to the location of the go-ipfs binary
     let go_ipfs_path = env::vars()
         .find(|(key, _val)| key == "GO_IPFS_PATH")
-        .expect("the GO_IPFS_PATH environment variable not found")
+        .expect("the GO_IPFS_PATH environment variable was not found")
         .1;
 
     let mut tmp_dir = env::temp_dir();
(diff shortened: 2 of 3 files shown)


ljedrz:
the conformance test failure is unrelated.

src/lib.rs
```diff
@@ -116,7 +116,7 @@ impl IpfsOptions {
             keypair: Keypair::generate_ed25519(),
             mdns: Default::default(),
             bootstrap: Default::default(),
-            kad_protocol: Default::default(),
+            kad_protocol: Some("/ipfs/lan/kad/1.0.0".to_owned()),
```
koivunej, line 119:
```suggestion
            // default to lan kad for go-ipfs use in tests
            kad_protocol: Some("/ipfs/lan/kad/1.0.0".to_owned()),
```

or something like this? I at least will forget the reason for in about 60 seconds :)

tests/kademlia.rs
```diff
@@ -60,102 +101,178 @@ async fn find_peer_dht() {
...
 
-    assert_eq!(found_addrs, vec![ids_and_addrs[CHAIN_LEN - 1].1.clone()]);
+    (nodes, ids_and_addrs, None)
 }
 
-// TODO: split into separate, more advanced bootstrap and closest_peers
-#[tokio::test(max_threads = 1)]
-async fn kademlia_local_peer_discovery() {
-    const BOOTSTRAPPER_COUNT: usize = 20;
+#[cfg(feature = "test_dht_with_go")]
```
koivunej, line 127: There shouldn't be any issue running this test on CI apart from some scripting, did you consider that? I think we should have curl, tar and friends available on all three platforms.

ljedrz: I did, just wanted to make this self-contained first and add a CI entry as a follow-up

tests/kademlia.rs
```diff
@@ -60,102 +101,178 @@ async fn find_peer_dht() {
...
+fn start_go_node() -> GoIpfsNode {
+    // GO_IPFS_PATH should point to the location of the go-ipfs binary
+    let go_ipfs_path = env::vars()
+        .find(|(key, _val)| key == "GO_IPFS_PATH")
+        .expect("the GO_IPFS_PATH environment variable not found")
```
ljedrz, line 132:
```suggestion
        .expect("the GO_IPFS_PATH environment variable was not found")
```

ljedrz:
rebased for a better review experience

tests/kademlia.rs
koivunej: Good good

koivunej (approved):
bors r+

PR #348, opened 2020-08-30
pass another dht.query conformance test

ljedrz:
A small fix that allows us to pass one more `dht` conformance test, namely `"should return the other node in the query"`.


koivunej (approved):
bors r+

PR #353, opened 2020-09-01
Implement /dns and /resolve

ljedrz:
Add a rudimentary implementation of the `/dns` and `/resolve` endpoints; putting it out there already, as due to the similarity of these two endpoints I'm not 100% sure how much we want to "condense" their inner workings.

This upgrades or conformance suite stats from
```
170 passing
50 pending
```
to
```
178 passing
42 pending
```

### DIFF
--- a/src/ipns/dns.rs
+++ b/src/ipns/dns.rs
@@ -18,7 +18,7 @@ use thiserror::Error;
 #[error("no dnslink entry")]
 pub struct DnsLinkError;
 
-type FutureAnswer = Pin<Box<dyn Future<Output = Result<Answer, io::Error>>>>;
+type FutureAnswer = Pin<Box<dyn Future<Output = Result<Answer, io::Error>> + Send>>;
 
 pub struct DnsLinkFuture {
     query: SelectOk<FutureAnswer>,
--- a/src/ipns/dns.rs
+++ b/src/ipns/dns.rs
@@ -8,15 +8,21 @@ use domain_resolv::{stub::Answer, StubResolver};
 use futures::future::{select_ok, SelectOk};
 use futures::pin_mut;
 use std::future::Future;
-use std::io;
 use std::pin::Pin;
 use std::str::FromStr;
 use std::task::{Context, Poll};
-use thiserror::Error;
+use std::{fmt, io};
 
-#[derive(Debug, Error)]
-#[error("no dnslink entry")]
-pub struct DnsLinkError;
+#[derive(Debug)]
+pub struct DnsLinkError(String);
+
+impl fmt::Display for DnsLinkError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(fmt, "DNS error: {}", self.0)
+    }
+}
+
+impl std::error::Error for DnsLinkError {}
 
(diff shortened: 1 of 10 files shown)


src/ipns/dns.rs
```diff
@@ -18,7 +18,7 @@ use thiserror::Error;
 #[error("no dnslink entry")]
 pub struct DnsLinkError;
 
-type FutureAnswer = Pin<Box<dyn Future<Output = Result<Answer, io::Error>>>>;
```
ljedrz, line 21: note: this one gave me a bit of a headache - warp was pointing to a **very** long issue with the HTTP call, but this was actually the root cause of the problem

ljedrz:
seems that I've missed a test - I'll leave this one here as a draft and tinker with it some more

ljedrz:
the curious thing is that only the `/dns` tests fail on Windows, while the `/resolve` ones pass, even though they both use the same internal machinery :thinking:; why does Windows not resolve `ipfs.io`?

src/lib.rs
```diff
@@ -446,11 +446,26 @@ impl<Types: IpfsTypes> Ipfs<Types> {
...
+                    }
+                }
+                previous = Some(res.clone());
+                resolved = ipns.resolve(&res).await;
+            }
+
+            resolved
+        } else {
+            resolved
+        }
```
koivunej, line 468: could add the:

```
async move {

}.instrument(self.span.clone()).await
```

around this to get back the instrumentation. This probably will make our compilation times grow towards infinity but seems to work :)

src/lib.rs
```diff
@@ -446,11 +446,26 @@ impl<Types: IpfsTypes> Ipfs<Types> {
...
+        if recursive {
+            let mut previous = None;
+            while let Ok(ref res) = resolved {
+                if let Some(ref prev) = previous {
+                    if prev == res {
+                        break;
+                    }
+                }
+                previous = Some(res.clone());
+                resolved = ipns.resolve(&res).await;
```
koivunej, line 462: I think here should be a hashmap of seen results. I can at least imagine it can be easy to create a loop with two domains.

src/ipns/dns.rs
koivunej: Yeah I wonder if the future should be just written open as `async fn` to be simpler in every way, but haven't tried this myself ... for a while at least. Though, there's nothing obvious causing this, perhaps the path inside domain-resolve is just so long.

src/lib.rs
ljedrz: ah yeah, I forgot about this :+1:

src/lib.rs
ljedrz: good point :+1:

aphelionz (approved):
bors r+

ljedrz:
in the end the issue was that the `domain_resolv` crate just doesn't read the DNS server addresses in Windows; [the docs for `ResolvConf::default`](https://docs.rs/domain-resolv/0.5.1/domain_resolv/stub/conf/struct.ResolvConf.html#method.default) (which is used by the `StubResolver`) state:
```
XXX This currently only works for Unix-y systems.
```
the solution was to use the `ipconfig` crate in order to read the DNS configuration in Windows and to provide that information to the `StubResolver` manually.

ljedrz:
Cc https://github.com/rs-ipfs/rust-ipfs/issues/6

PR #354, opened 2020-09-02
add DHT get and put

ljedrz:
While I'm looking at the curious Windows crashes with `/dns` and `/resolve`, we can expand territories elsewhere: I've had these functionalities implemented in a branch building one the recently merged PRs; there's no counterpart HTTP endpoints yet, as the conformance tests use functionalities that we're not implementing yet; these features are less interesting for the HTTP API anyway.

### DIFF
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -525,15 +525,16 @@ impl<Types: IpfsTypes> Behaviour<Types> {
         }
     }
 
-    pub fn dht_get(&mut self, key: Key) -> SubscriptionFuture<KadResult, String> {
+    pub fn dht_get(&mut self, key: Key, quorum: Quorum) -> SubscriptionFuture<KadResult, String> {
         self.kad_subscriptions
-            .create_subscription(self.kademlia.get_record(&key, Quorum::One).into(), None)
+            .create_subscription(self.kademlia.get_record(&key, quorum).into(), None)
     }
 
     pub fn dht_put(
         &mut self,
         key: Key,
         value: Vec<u8>,
+        quorum: Quorum,
     ) -> Result<SubscriptionFuture<KadResult, String>, anyhow::Error> {
         let record = Record {
             key,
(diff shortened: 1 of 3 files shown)


src/p2p/behaviour.rs
```diff
@@ -515,6 +524,31 @@ impl<Types: IpfsTypes> Behaviour<Types> {
             }
         }
     }
+
+    pub fn dht_get(&mut self, key: Key) -> SubscriptionFuture<KadResult, String> {
+        self.kad_subscriptions
+            .create_subscription(self.kademlia.get_record(&key, Quorum::One).into(), None)
```
koivunej, line 530: setting the quorum to one seems a bit strange choice but probably helps passing the tests.

Should `K` be used here instead, maybe a FIXME or TODO comment? The HTTP API docs for these methods does not at least mention being able to specify the quorum. At least looking at `ipfs dht get --help` it discusses finding multiple values instead of the first and selecting the "best" where "best" is defined only for ipns records.

src/p2p/behaviour.rs
```diff
@@ -515,6 +524,31 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
+        key: Key,
+        value: Vec<u8>,
+    ) -> Result<SubscriptionFuture<KadResult, String>, anyhow::Error> {
+        let record = Record {
+            key,
+            value,
+            publisher: None,
+            expires: None,
+        };
+        match self.kademlia.put_record(record, Quorum::One) {
```
koivunej, line 544: Similar quorum question here.

koivunej (commented):
Questions re: quorum and at concerns re: commenting at least.

src/p2p/behaviour.rs
ljedrz: good point :+1:; I left this decision to the caller for now

koivunej (approved):
bors r+

Issue #173, opened 2020-05-25
Configurable listening addresses


koivunej:
Currently rust-ipfs defaults to listening on a random localhost port like many rust-libp2p examples. In #169 the initial support for adding and removing listening addresses was implemented to realize a test case of a peer having multiple connections but the old behaviour of listening on a random localhost address remained.

rust-ipfs could however take the listening address(es) as a `Multiaddr` in `IpfsOptions`. This would allow removal of the random listening address.

This option could be populated per config section "Addresses" by the http.

PR #364, opened 2020-09-09
Configurable listening addrs

ljedrz:
This PR allows us to specify the desired listening address(es) of the node. For now, these addresses are just additions to the current randomly assigned listening address, as I'm not sure what to pick as the default; once we know this, the only change needed (besides the default listening addresses being put in the config) is to remove the following line from `p2p::create_swarm`:
```
// Listen on all interfaces and whatever port the OS assigns
Swarm::listen_on(&mut swarm, "/ip4/127.0.0.1/tcp/0".parse().unwrap()).unwrap();
```
Another question is how to handle a situation where one/all of the supplied addresses can't be listened on:
- refuse to start?
- assign a random different address?
- only log errors if at least one works?

This is currently unresolved, but since the PR preserves the current behavior, this is a concern that needs to be handled in the aforementioned later follow-up.

Cc https://github.com/rs-ipfs/rust-ipfs/issues/173.

### DIFF
--- a/http/src/main.rs
+++ b/http/src/main.rs
@@ -137,7 +137,8 @@ fn main() {
     let mut rt = tokio::runtime::Runtime::new().expect("Failed to create event loop");
 
     rt.block_on(async move {
-        let opts: IpfsOptions = IpfsOptions::new(home.clone(), keypair, Vec::new(), false, None);
+        let opts: IpfsOptions =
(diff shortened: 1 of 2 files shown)


aphelionz (approved):
Very nice!

bors r+

ljedrz:
The conformance test failure is unrelated.
bors retry

PR #365, opened 2020-09-09
Improve multi-node tests

ljedrz:
This PR improves our multi-node testing capabilities by:
- introducing a `spawn_connected_nodes` function and the related `Topology` object, allowing us to automate the creation of various test scenarios
- adds the `PeerId` and listening addresses to the `Node` object used in tests, to remove the need for calling `.identity()` every time a node is created

These new functionalities are not used everywhere yet and can probably be improved further, but it's a good starting point.

### DIFF
--- a/src/dag.rs
+++ b/src/dag.rs
@@ -596,7 +596,7 @@ mod tests {
 
     #[tokio::test(max_threads = 1)]
     async fn test_resolve_root_cid() {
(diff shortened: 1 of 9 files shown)


ljedrz:
seems that the `wait` that I removed was a bit too optimistic; I'll bring it back to that one test

aphelionz (approved):
LGTM

bors r+

PR #366, opened 2020-09-10
Finish Block subscriptions only if a subscription had been created

ljedrz:
I've found this issue while doing runs of the usually `#[ignore]`d tests, namely with the `bitswap_stress_test`. In `debug` builds it would trigger a `debug_assert` that verifies that we don't finish non-existent subscriptions (which in itself is not a big deal, but it is undesirable).

The reason was that a call to `get_block` that doesn't find the `Block` in the local repo creates a subscription for that `Block` (but _not_ if it was found), while `put_block` always finished a subscription for the `Block`, even if it had already existed in the repo. The solution was to just move `finish_subscription` under the `if let BlockPut::NewBlock = res` block, meaning that it's triggered only if the block that was put was new to the repo.


aphelionz (approved):
bors r+

PR #367, opened 2020-09-10
Reorganize some tests

ljedrz:
The individual commits describe specific changes; it's mostly renaming some test files and moving a few tests around, but also the removal of some dead tests.

I'd like to push these changes further and move the tests submodules from `lib`, `dag`, and `refs` to `tests`, which would make those modules smaller and the tests more easily accessible, but also allow us to move the `Node` object from `lib` to `tests::common`. The reason I'm not doing it yet is that it would require some currently private functions and objects to become public and while I could just do that and "slap" `#[doc(hidden)]` on top of them, I'd prefer @koivunej to also have a say in how to tackle this.


aphelionz (approved):
bors r+

PR #371, opened 2020-09-15
Partial /bootstrap support

ljedrz:
Somewhat contrary to this endpoint's name, it doesn't cause any bootstrapping to happen, just modifies the bootstrapper node list. I thought I might make all the conformance tests pass before publishing this, but I then noticed that it's already a pretty big PR, so I decided to push it already. The commits, while not exactly the smallest, are more or less self-contained and organized in a sensible order.

The change in conformance test suite score is currently an extra `9 passing, 6 pending`; actually 10 new tests now pass, but one of them has the same description as a failing one, so both get disabled. I hope to have the failing ones working soon.


src/p2p/behaviour.rs
```diff
@@ -550,6 +556,72 @@ impl<Types: IpfsTypes> Behaviour<Types> {
...
+        self.swarm.bootstrappers.remove(&addr);
+        Ok(ret)
+    }
+
+    pub fn clear_bootstrappers(&mut self) -> Vec<Multiaddr> {
+        self.swarm.bootstrappers.drain().map(|a| a.into()).collect()
+    }
+
+    pub fn restore_bootstrappers(&mut self) -> Vec<Multiaddr> {
+        let config_location = std::env::var("IPFS_PATH").expect("always expected to be set");
```
ljedrz, line 597: there's an error in the path here, which is already fixed in the incoming follow-up; also, the `expect`s, however unlikely, are made into nicer errors

aphelionz (approved):
bors r+

ljedrz:
unrelated conformance test failure
bors retry

PR #372, opened 2020-09-15
Full /bootstrap support

ljedrz:
~~Builds on https://github.com/rs-ipfs/rust-ipfs/pull/371, only the last 6 commits are new.~~

Fixes the remaining 5 `bootstrap` conformance tests, bringing our overall score past 200 passing :muscle::
```
201 passing (52s)
48 pending
```

As usual, the commits describe specific changes; most are pretty self-explanatory, with the exception of the new patch:
```
-      expect(peers).to.have.property('length').that.is.gt(1)
+      expect(peers).to.have.property('length').that.is.eq(0)
```
The reason behind that one is that our HTTP test node only carries `Identity` config information - there is no `Bootstrap` key available in the config file, so the number of default bootstrapper addresses is `0`.

~~Blocked by https://github.com/rs-ipfs/rust-ipfs/pull/371.~~


aphelionz (approved):
bors r+

PR #374, opened 2020-09-16
Fix 2 issues triggered with the dht_popular_content_discovery test

ljedrz:
This is an `#[ignore]`d test due to it using the global network, but it's a practical debugging tool regardless; this time it helped fix 2 issues:
- the handling of `Multiaddr`s ending with `/p2p-circuit`
- searching for `Block` providers using the full `Cid` instead of just the `Multiaddr` (we've noticed this before, but it got lost in the TODOs :stuck_out_tongue:)


aphelionz (approved):
bors r+

PR #375, opened 2020-09-16
A fistful of tweaks

ljedrz:
A pretty wide range of test improvements, small fixes, refactorings and renames. Each commit describes the specific change, but I'll list the more noteworthy ones:
- automatically remove `Protocol::P2p(_)` from the `Multiaddr`s passed to `Ipfs::add_peer`
- only produce a `RepoEvent::RemovedBlock` when a block was actually removed
- add unit tests for the `Node` topologies


aphelionz (approved):
LTGM

bors r+

PR #377, opened 2020-09-17
Enabling pinned blocks test

aphelionz:
This PR un-skips one conformance test that works now that we have pinning. This brings our number up by one! :firecracker:


ljedrz (approved):
Throwing errors on attempts to remove pinned blocks is already supported, so LGTM.
bors r+

PR #378, opened 2020-09-17
Interop changes

ljedrz:
While I was looking at the upcoming `go-ipfs 0.7` changes, the one that stood out to me most was the switch from RSA to Ed25519 keys by default; since the only reason why we're shipping `openssl` is that we be default generate RSA keys, this seemed to me like a good opportunity to ditch it and free ourselves from one of the trickiest dependencies.

Since I might have been a bit "heavy-handed" in the way I handled these changes and since they are quite severe, I'm posting it as a draft so that we can discuss this further, see how the CI changes fare etc.


ljedrz:
Nice, the CI seems to agree with this approach; I'll just force-push the removal of `libssl-dev` I forgot to remove initially.

PR #380, opened 2020-09-18
Improve /add interop

ljedrz:
I browsed the `go-ipfs` and `js-ipfs` codebases in search for some nice `/add` test cases so that we could verify that our logic, e.g. wrt. the headers and field names, is compatible; I haven't found much :sweat_smile:, but based on the one `js-ipfs` test case I found I relaxed the `/add` restrictions a little bit.

Not sure if this is the right approach and if I'm seeing the whole picture, so I'm marking this as a draft.

### DIFF
--- a/http/src/v0/root_files/add.rs
+++ b/http/src/v0/root_files/add.rs
@@ -143,15 +143,7 @@ where
             let content_type = field.content_type().map_err(AddError::Header)?;
 
             let next = match content_type {
-                "application/octet-stream" => {
-
-                    // files are of the form "file-{1,2,3,..}"
-                    let _ = if field_name != "file" && !field_name.starts_with("file-") {
-                        Err(AddError::UnsupportedField(field_name.to_string()))
-                    } else {
-                        Ok(())
-                    }?;
-
+                "application/octet-stream" | "text/plain" => {
                     let mut adder = FileAdder::default();
                     // how many bytes we have stored as blocks
                     let mut total_written = 0u64;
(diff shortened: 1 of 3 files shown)


http/src/v0/root_files/add.rs
```diff
@@ -408,15 +404,37 @@ mod tests {
...
+                        .collect::<String>()
+                )
+                .into_bytes(),
+            )
+            .reply(&add(&ipfs))
+            .await;
 
-        tokio::spawn(fut);
-        ipfs
+        assert_eq!(response.status(), 200);
```
koivunej, line 438: is this test case is just to accept the text/plain as I guess js-ipfs accepts it?

accepting text/plain sounds really strange be honest ... wouldn't that require some charset wrangling as well? I guess there's no issues if you assume ascii or utf8 but ... Why?

http/src/v0/root_files/add.rs
```diff
@@ -143,15 +143,7 @@ where
...
 
             let next = match content_type {
-                "application/octet-stream" => {
-
-                    // files are of the form "file-{1,2,3,..}"
-                    let _ = if field_name != "file" && !field_name.starts_with("file-") {
-                        Err(AddError::UnsupportedField(field_name.to_string()))
-                    } else {
-                        Ok(())
-                    }?;
```
koivunej, line 153: not sure if this is a good idea to be removed. I mean if you think of a html form which is pretty much the basis for http multipart and fields with different names kinda should be different things so I'd like to retain some validation OR targeting here. Or are the go-ipfs and js-ipfs using random field names here? I'd rather see them "enumerated" here.

my comment is a bit off here, I think js-ipfs-http-client does the file-N.

koivunej (changes requested):
comments

http/src/v0/root_files/add.rs
ljedrz: this was the way `js-ipfs` did it, I just followed suit; since it's random alphanumeric bytes there's no encoding issues

PR #381, opened 2020-09-18
Add the list of supported libp2p protocols to /id

ljedrz:
Following the changes in [`go-ipfs`](https://github.com/ipfs/go-ipfs/pull/7409) and [`js-ipfs`](https://github.com/ipfs/js-ipfs/pull/3250) (and already required in the upstream conformance tests), this PR adds the quite useful list of supported `libp2p` protocols to the JSON returned by `/id`.

Since I'm not sure about the resulting small API changes and the way the protocols get populated (which was WAY harder than anticipated), I'm marking it as a draft.

### DIFF
--- a/src/p2p/swarm.rs
+++ b/src/p2p/swarm.rs
@@ -46,6 +46,7 @@ pub struct SwarmApi {
     roundtrip_times: HashMap<PeerId, Duration>,
     connected_peers: HashMap<PeerId, Vec<MultiaddrWithoutPeerId>>,
     pub(crate) bootstrappers: HashSet<MultiaddrWithPeerId>,
+    pub(crate) protocols: Vec<Vec<u8>>,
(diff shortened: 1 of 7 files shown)


src/p2p/swarm.rs
```diff
@@ -242,8 +243,12 @@ impl NetworkBehaviour for SwarmApi {
     fn poll(
         &mut self,
         _: &mut Context,
-        _: &mut impl PollParameters,
+        params: &mut impl PollParameters,
     ) -> Poll<NetworkBehaviourAction> {
+        if params.supported_protocols().len() != self.protocols.len() {
```
ljedrz, line 248: `PollParameters::supported_protocols` returns an `ExactSizeIterator`, so obtaining the `len` is cheap

PR #382, opened 2020-09-21
return errors for some extra DHT issues

ljedrz:
Return explicit subscription errors on some of the more user-facing DHT requests, namely `Bootstrap`, `StartProviding` and `PutRecord`; `Kademlia` errors for those are currently only logged and `KadResult::Complete` is returned, potentially concealing issues.

### DIFF
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -85,7 +85,9 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<KademliaEvent> for Behaviour
                     match result {
                         // these subscriptions return actual values
                         GetClosestPeers(_) | GetProviders(_) | GetRecord(_) => {}
-                        // and the rest just a general KadResult::Complete
+                        // we want to return specific errors for the following
+                        Bootstrap(Err(_)) | StartProviding(Err(_)) | PutRecord(Err(_)) => {}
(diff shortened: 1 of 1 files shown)


src/p2p/behaviour.rs
```diff
@@ -104,7 +106,14 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<KademliaEvent> for Behaviour
                         );
                     }
                     Bootstrap(Err(BootstrapError::Timeout { .. })) => {
-                        warn!("kad: failed to bootstrap");
+                        warn!("kad: timed out while trying to bootstrap");
+
+                        if self.kademlia.query(&id).is_none() {
```
koivunej, line 111: Just to clarify, this:

```
if self.kademlia.query(&id).is_none() {
```

Is done to to guard against the query returning a result but still being ongoing or something?

koivunej (approved):
I am not sure how good idea it's to always log an error before propagating it but perhaps here is no other choice, since we don't really know if the other side will log it, or if there is even something ongoing.

But yeah, seems that things are only moving towards better errors and more straightforward processing so approve.

src/p2p/behaviour.rs
ljedrz: yes, exactly that - some queries return partial results, and we don't want to finish the subscription until it's complete

Issue #356, opened 2020-09-03
Release 0.2.0


koivunej:
It's been literally forever since the last release and we've been discussing this internally. Releasing the crates will allow us to jump into making the docs better through docs.rs visibility and hopefully make the project more accessible through that way as well.

Thinking it's probably best not to even attempt at many last minute polish since we still have a lot of work ahead of us.

- [x] update bootstrap list to match latest [go-ipfs bootstrap peers](https://github.com/ipfs/go-ipfs-config/blob/master/bootstrap_peers.go#L17) #387
   - ~needs decision: resolve dnsaddr outside of libp2p, expand it manually or~ just not include /dnsaddr
- [x] use static bootstrap peers instead of [reloading config file](https://github.com/rs-ipfs/rust-ipfs/blob/ffc65c8b5de3390e94864904e11cbc425a3912e7/src/p2p/behaviour.rs#L596-L625) #387
- [x] ipfs-http: listening addresses from compat config file (see #173) #383
- [x] remove default random listening address, listen to no addresses by default (re: #173) #383
- [x] remove ipns insertion and cancellation features from ipfs::Ipfs (they are not implemented) #384
- [x] comment on resolve_ipns that it only supports dnslink was #385 but fused into #384
- [x] move README.md example as an example under `examples/`, stop polluting src/lib.rs docs #391 
- [x] be on a look out for other half-way implemented features and update docs #392 
- [x] crate level ipfs-http documentation (in #392)
- [x] update #6 regarding IPNS progress
- [x] go through remaining examples (removed the ipns one in #384) -- #390
- [x] simplify node creation #345 maybe? in #394
- [x] remove UninitializedIpfs::default, IpfsOptions::default and `src/config.rs` except for bootstrapper list, see [comment](https://github.com/rs-ipfs/rust-ipfs/issues/356#issuecomment-696780366), #393
- [x] ~add add example~, discussed more in deferral #395
- [x] crates.io prep, me as ipfs owner
- [x] ipfs-unixfs changelog, bump ipfs-unixfs to 0.2.0, #396
- [x] ipfs-unixfs tag+publish
- [x] prepare to publish `bitswap/` as ipfs-bitswap 0.1.0 #397 #398 (again)
- [x] tag+publish ipfs-bitswap
- [x] prepare to publish ipfs as 0.2.0 #399
- [x] tag+publish ipfs
- [x] create persistence tracking issue, current state #401
[...]

PR #383, opened 2020-09-22
remove the default random listening address

ljedrz:
Remove the default random listening address, listen to no addresses by default.

### DIFF
--- a/src/p2p/mod.rs
+++ b/src/p2p/mod.rs
@@ -59,12 +59,9 @@ pub async fn create_swarm<TIpfsTypes: IpfsTypes>(
     let behaviour = behaviour::build_behaviour(options, ipfs).await;
 
     // Create a Swarm
-    let mut swarm = libp2p::swarm::SwarmBuilder::new(transport, behaviour, peer_id)
+    let swarm = libp2p::swarm::SwarmBuilder::new(transport, behaviour, peer_id)
         .executor(Box::new(SpannedExecutor(swarm_span)))
         .build();
-
-    // Listen on all interfaces and whatever port the OS assigns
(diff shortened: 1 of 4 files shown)


src/p2p/mod.rs
```diff
@@ -59,13 +59,10 @@ pub async fn create_swarm<TIpfsTypes: IpfsTypes>(
     let behaviour = behaviour::build_behaviour(options, ipfs).await;
 
     // Create a Swarm
-    let mut swarm = libp2p::swarm::SwarmBuilder::new(transport, behaviour, peer_id)
+    let swarm = libp2p::swarm::SwarmBuilder::new(transport, behaviour, peer_id)
         .executor(Box::new(SpannedExecutor(swarm_span)))
         .build();
 
-    // Listen on all interfaces and whatever port the OS assigns
-    Swarm::listen_on(&mut swarm, "/ip4/127.0.0.1/tcp/0".parse().unwrap()).unwrap();
```
koivunej, line 67: the IpfsOptions factory fn used in tests needs to add this.

koivunej (changes requested):
https://github.com/rs-ipfs/rust-ipfs/blob/ffc65c8b5de3390e94864904e11cbc425a3912e7/src/lib.rs#L135 or `IpfsOptions::inmemory_with_generated_keys` needs to retain that random listening address.

ljedrz:
Hmm, the conformance tests need to be adjusted too, either at `http` config level when `profile = test` or in `index.js`.

ljedrz:
In the end the `ipfs-http` config file needed to be enhanced with the random listening address; the format is compatible with the `go-ipfs` config file.

koivunej (approved):
Good to go, only approving since I am not sure if there are some ordering stuff.

PR #384, opened 2020-09-22
remove unimplemented ipns functionalities from Ipfs

ljedrz:
Remove ipns insertion and cancellation features from `ipfs::Ipfs` (they are not implemented).


koivunej (changes requested):
So this would leave us:
- I see the `ipfs::Ipns::resolve` is different from `Ipfs::resolve_ipns`
- insert and cancel are removed, `src/ipns/mod.rs` becomes unnecessary except for `dns`
- `ipfs::ipns::dns` should probably be `dnslink`?
- build.rs compiles `src/ipns/ipns_pb.proto` for no reason, could be removed, but maybe the file be retained?
- example is now worthless, could be removed

Perhaps the `src/ipns/mod.rs` should be cleaned up to contain only comment on why it was removed? Or then just retain the current structure with `struct Ipns` and move the resolving method from `Ipfs::resolve_ipns` to it.

koivunej (approved):
Looking good, not merging yet, probably no deps but you know better.

PR #387, opened 2020-09-22
fix: restore bootstrappers from static list, update the bootstrapper list

ljedrz:
Alters the `BOOTSTRAP_NODES` list to contain only the currently working addresses and changes the behavior of `Swarm::remove_bootstrappers` to not use the config file, but rather the aforementioned in-memory list.

### DIFF
--- a/conformance/patches/bootstrap.patch
+++ b/conformance/patches/bootstrap.patch
@@ -5,7 +5,7 @@
 
        const peers = res.Peers
 -      expect(peers).to.have.property('length').that.is.gt(1)
-+      expect(peers).to.have.property('length').that.is.eq(0)
++      expect(peers).to.have.property('length').that.is.gt(0)
      })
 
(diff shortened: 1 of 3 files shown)


koivunej (changes requested):
The conformance test is because of the patched test case, so the patch can now go?

ljedrz:
I guess so; I'll duble-check.

ljedrz:
Ok, so it needs to remain, but be adjusted to:
```
expect(peers).to.have.property('length').that.is.gt(0) // as opposed to .gt(1)
```
because we currently ship with just a single bootstrapper address.

koivunej (approved):
Allrighty... Sounds like to me this should be fixed upstream as well. Not sure what the aim of testing `bootstrappers.length >= 1` they cannot already confirm with `bootstrappers.length > 1`. I mean, if it was two, might as well assert that the two are found in the list.

koivunej:
Whoops

bors r+

Issue #197, opened 2020-06-18
Documentation for "Public" Methods


aphelionz:
Rust documentation can be improved overally, but mainly in two places:

1. Instantiating and setting up an IPFS node
2. All of the methods in the `Ipfs` struct, i.e. the main functions that a Rust programmer will use to interact with an IPFS node

koivunej:
After #392 we are now about 60% documented according to docs.rs.

If you are interested in helping out, probably easiest way to find anything to work on is to either:
* clone, commend out [src/lib.rs:18](https://github.com/rs-ipfs/rust-ipfs/blob/722d41a0864148e15721e1bca39e045ca13ce510/src/lib.rs#L18), run `cargo check`
  * this will produce around 180 errors but works well with rust-analyzer and an IDE
* browse [docs.rs/ipfs](https://docs.rs/ipfs)

koivunej:
Updated the issue description with the nightly/docs.rs hints, lint recommendations.

PR #409, opened 2020-10-06
Improve documentation of public methods 

niklaslong:
This PR goes some way towards improving the documentation of public methods as discussed in #197.

### DIFF
--- a/src/p2p/addr.rs
+++ b/src/p2p/addr.rs
@@ -82,7 +82,9 @@ impl FromStr for MultiaddrWithoutPeerId {
 /// don't support it being contained within the `Multiaddr`.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct MultiaddrWithPeerId {
+    /// The peer without its address.
     pub multiaddr: MultiaddrWithoutPeerId,
+    /// The address of the peer.
     pub peer_id: PeerId,
 }
 
--- a/src/p2p/addr.rs
+++ b/src/p2p/addr.rs
@@ -82,9 +82,9 @@ impl FromStr for MultiaddrWithoutPeerId {
 /// don't support it being contained within the `Multiaddr`.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct MultiaddrWithPeerId {
-    /// The peer without its address.
+    /// The [`Multiaddr`] without the [`Protocol::P2p`] suffix.
     pub multiaddr: MultiaddrWithoutPeerId,
-    /// The address of the peer.
+    /// The peer id from the [`Protocol::P2p`] suffix.
     pub peer_id: PeerId,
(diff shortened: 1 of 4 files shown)


src/p2p/mod.rs
```diff
@@ -17,13 +17,20 @@ mod transport;
 pub use addr::{MultiaddrWithPeerId, MultiaddrWithoutPeerId};
 pub use {behaviour::KadResult, swarm::Connection};
 
+/// An IPFS swarm.
```
koivunej, line 20:
```suggestion
/// Type alias for [`libp2p::Swarm`] running the [`behaviour::Behaviour`] with the given [`IpfsTypes`].
```

Not sure if "running" is the correct verb here. All in all this type alias might be a bit unnecessary.

src/p2p/mod.rs
```diff
@@ -17,13 +17,20 @@ mod transport;
...
 pub struct SwarmOptions {
+    /// The keypair for the PKI based identity of the local node.
     pub keypair: Keypair,
+    /// The peer address of the local node created from the keypair.
     pub peer_id: PeerId,
+    /// The peers to connect to on startup.
     pub bootstrap: Vec<(Multiaddr, PeerId)>,
+    /// Enables mdns for peer discovery and announcement when true.
     pub mdns: bool,
+    /// Custom Kademlia protocol name.
```
koivunej, line 33:
```suggestion
    /// Custom Kademlia protocol name, see [`IpfsOptions::kad_protocol`].
```
I didn't check if that'll be a good link however linking to the longer description with the default in `IpfsOptions` seems like the way to go. Another way would be to link to `IpfsOptions` at the type level.

src/p2p/addr.rs
```diff
@@ -82,7 +82,9 @@ impl FromStr for MultiaddrWithoutPeerId {
 /// don't support it being contained within the `Multiaddr`.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct MultiaddrWithPeerId {
+    /// The peer without its address.
```
koivunej, line 85:
```suggestion
    /// The [`Multiaddr`] without the [`Protocol::P2p`] suffix.
```

src/p2p/addr.rs
```diff
@@ -82,7 +82,9 @@ impl FromStr for MultiaddrWithoutPeerId {
...
     pub multiaddr: MultiaddrWithoutPeerId,
+    /// The address of the peer.
```
koivunej, line 87:
```suggestion
    /// The peer id from the [`Protocol::P2p`] suffix.
```

koivunej (approved):
Few small adjustments and we are ready to go. After you've handled the suggestions as you see fit you can ask bors to merge this by `bors r+` on a single line in comment.

bors d+

PR #412, opened 2020-10-06
doc(unixfs): fix typos

c410-f3r:


koivunej (approved):
Hi @c410-f3r and thanks for reading through :) No changelog addition needed but I did adjust the PR subject so that I'll remember faster when going over and making sure nothing was lost.

bors r+

PR #415, opened 2020-10-13
add unixfs benchmark

niklaslong:
Opening this as a draft to have a discussion around the performance of the unixfs adder. 

Things I've observed so far: 
- benchmark is a little noisy (on my machine at least)
- the adder is significantly slower and more unstable when chunking by small amounts, though this is expected
- I'm not seeing clear spikes for factors of 174, just a general trend towards lower throughput as the input gets larger
- black boxing the returned blocks didn't seem to have much of an impact (small chunking size) 

I'm not sure what the best next step would be; as this is meant as a base for future improvements to unixfs, what values/resolution do we want for this?


unixfs/benches/adder.rs
```diff
@@ -0,0 +1,84 @@
...
+//      2.295 MB in 15.00s or 0.070 MBps
+//      2.365 MB in 16.00s or 0.070 MBps
+//      2.435 MB in 17.00s or 0.070 MBps
+//      2.502 MB in 18.00s or 0.067 MBps
+//      2.567 MB in 19.00s or 0.065 MBps
+//      2.630 MB in 20.00s or 0.063 MBps
+//      2.693 MB in 21.00s or 0.063 MBps
+//      2.752 MB in 22.00s or 0.060 MBps
+//      2.811 MB in 23.00s or 0.059 MBps
+//      2.869 MB in 24.00s or 0.058 MBps
```
koivunej, line 45: it would appear that my test had bad formatting, the MBps is during the last second, not the average as it might read like. `s/or/at/` but this is ok.

koivunej (approved):
bors d+

Issue #402, opened 2020-09-24
ipfs-http: manual testability or discoverability


koivunej:
This was cut from [the v0.2.0 release](https://github.com/rs-ipfs/rust-ipfs/issues/356) for present time constraints however making ipfs-http more accessible as "the" ipfs binary would make the project more accessible to anyone just wanting to try it out. At the moment I see the following obstacles:

- `http/README.md` has no tutorial on how to try out `ipfs-http`
- `ipfs-http` supports only `init --profile test --bits b` where b >= 2048
- `ipfs-http` only binds to ephemeral localhost port for http
- `ipfs-http` currently by default uses `~/.ipfs` which is go-ipfs uses
- `ipfs-http` is "a bit incompatible" with go-ipfs 0.6 CLI at least, which #380 started to explore
- there is no repo locking to prevent trying to connect two peers with same ids: #243 

The tutorial could be simple one which would show how to interact using go-ipfs CLI with the `ipfs-http`, something along the lines of:

1. install rust (link to book)
2. clone repo
3. `cargo run -p ipfs-http -- init`
4. `cargo run -p ipfs-http -- daemon`
5. in other terminal: `IPFS_PATH=~/.rust-ipfs go-ipfs-binary-somewhere id`

`ipfs-http` supporting only the `init --profile test --bits b` where b >= 2048 would be quite easy to fix so that it would support `init` and `init --bits b` where b >= 2048 by adding a profile abstraction or even a simple if-tornado in place of: https://github.com/rs-ipfs/rust-ipfs/blob/722d41a0864148e15721e1bca39e045ca13ce510/http/src/config.rs#L68-L74

(I haven't really looked at go-ipfs init profiles, they might have more in-depth features but the "test or default" would most likely be enough here.)

The `ipfs::config::BOOTSTRAP_NODES` would need to be populated whenever in this "default" or not given profile along with listening address of `/ip4/0.0.0.0/tcp/4004`. Also the http API port would need to be specified as it is quite frustrating to copy and paste the multiaddr from the stdout produced by `cargo run -p ipfs-http -- daemon` and convert it to normal address for curl by hand.

The "a bit incompatible" can be seen by trying to `go-ipfs-0.6 add` stuff in; I was being too conservative here and go-ipfs probably uses just "file" as the field name: https://github.com/rs-ipfs/rust-ipfs/blob/master/http/src/v0/root_files/add.rs#L148-L153

[...]

PR #419, opened 2020-10-21
fix(http): change default ipfs path to .rust-ipfs

niklaslong:
This PR changes the default `IPFS_PATH` from `.ipfs` to `.rust-ipfs` as mentioned in #402. 

I _think_ this is all that's needed?


koivunej (approved):
Yeah it probably didn't need to be more complicated than this.

bors r+

PR #421, opened 2020-10-22
feat(http): create Profile abstraction to allow port choice

niklaslong:
This PR introduces a `Profile` abstraction to allow the customisation of port selection as discussed in #402.

Two profiles are supported: 
- `Test` for use with conformance tests (ephemeral port selection)
- `Default` serves on `4004`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Next
 
-* fix(http): create `Profile` abstraction [#421]
+* feat(http): create `Profile` abstraction [#421]
 
 [#421]: https://github.com/rs-ipfs/rust-ipfs/pull/421
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next
 
+* fix(http): create `Profile` abstraction [#421]
+
+[#421]: https://github.com/rs-ipfs/rust-ipfs/pull/421
+
 # 0.2.1
 
 * fix: restore_bootstrappers doesn't enable content discovery [#406]
--- a/http/src/config.rs
+++ b/http/src/config.rs
@@ -22,6 +22,7 @@ pub enum Profile {
     Default,
 }
 
+// Required for structopt.
 impl FromStr for Profile {
     type Err = InitializationError;
 
@@ -60,6 +61,16 @@ pub fn initialize(
     bits: NonZeroU16,
     profiles: Vec<Profile>,
 ) -> Result<(), InitializationError> {
+    // This check is done here to avoid an empty config file being created in the case of an
+    // unsupported input.
+    if profiles.len() != 1 {
+        // profiles are expected to be (comma separated) "test" as there are no bootstrap peer
+        // handling yet. the conformance test cases seem to init `go-ipfs` in this profile where
+        // it does not have any bootstrap nodes, and multi node tests later call swarm apis to
+        // dial the nodes together.
+        unimplemented!("Multiple profiles are currently unsupported!");
+    }
+
     let config_path = ipfs_path.join("config");
 
     fs::create_dir_all(&ipfs_path)
@@ -78,6 +89,11 @@ fn create(
     use multibase::Base::Base64Pad;
     use prost::Message;
     use std::io::BufWriter;
+
+    let api_addrs = match profiles[0] {
+        Profile::Test => "127.0.0.1:0",
+        Profile::Default => "127.0.0.1:4004",
+    };
 
     let bits = bits.get();
 
@@ -123,19 +139,6 @@ fn create(
     };
 
     let private_key = Base64Pad.encode(&private_key);
-
-    if profiles.len() != 1 {
-        // profiles are expected to be (comma separated) "test" as there are no bootstrap peer
-        // handling yet. the conformance test cases seem to init `go-ipfs` in this profile where
-        // it does not have any bootstrap nodes, and multi node tests later call swarm apis to
-        // dial the nodes together.
-        unimplemented!("Multiple profiles are currently unsupported!")
-    }
-
-    let api_addrs = match profiles[0] {
-        Profile::Test => "127.0.0.1:0",
-        Profile::Default => "127.0.0.1:4004",
-    };
 
     let config_contents = CompatibleConfigFile {
         identity: Identity {
--- a/http/src/config.rs
+++ b/http/src/config.rs
@@ -1,9 +1,8 @@
 //! go-ipfs compatible configuration file handling and setup.
 
-use parity_multiaddr::Multiaddr;
+use parity_multiaddr::{multiaddr, Multiaddr};
 use serde::{Deserialize, Serialize};
 use std::fs::{self, File};
-use std::net::SocketAddr;
 use std::num::NonZeroU16;
 use std::path::Path;
 use std::str::FromStr;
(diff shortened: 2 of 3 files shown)


http/src/config.rs
```diff
@@ -37,8 +59,18 @@ pub enum InitializationError {
...
-    profiles: Vec<String>,
+    profiles: Vec<Profile>,
 ) -> Result<(), InitializationError> {
+    // This check is done here to avoid an empty config file being created in the case of an
+    // unsupported input.
+    if profiles.len() != 1 {
+        // profiles are expected to be (comma separated) "test" as there are no bootstrap peer
+        // handling yet. the conformance test cases seem to init `go-ipfs` in this profile where
+        // it does not have any bootstrap nodes, and multi node tests later call swarm apis to
+        // dial the nodes together.
```
koivunej, line 70:
```suggestion
```

I think the comment is now mostly outdated; configuring bootstrap peers in the config file will be coming.

http/src/config.rs
```diff
@@ -37,8 +59,18 @@ pub enum InitializationError {
 pub fn initialize(
     ipfs_path: &Path,
     bits: NonZeroU16,
-    profiles: Vec<String>,
+    profiles: Vec<Profile>,
 ) -> Result<(), InitializationError> {
+    // This check is done here to avoid an empty config file being created in the case of an
+    // unsupported input.
```
koivunej, line 65:
```suggestion
```

I think this comment is covered by the `unimplemented!` and the following `profiles[0]` on the https://github.com/rs-ipfs/rust-ipfs/blob/4b58f0cdf77c1c6d5a277da8ccf26f6b8fe9da9f/http/src/config.rs#L93-L96 I don't think it's possible to fallthrough here and get an empty list for addresses as the `profiles[0]` would had panicked.

Given the panic and existing failure possibilities in `create` I am now thinking if the address (and most other parts of CompatibleConfigFile) should be produced by or with the help of the profiles before the file is even created. The existing structure of this simple config file creation does seem quite unfortunate..

Perhaps a more drastic restructuring is in order? A panic or failure from `create` seems to currently leave behind an empty file. A more drastic restructuring would create the in-memory CompatibleConfigFile fully and only serialize it as the last step.

http/src/config.rs
```diff
@@ -245,6 +275,7 @@ struct CompatibleConfigFile {
 #[serde(rename_all = "PascalCase")]
 struct Addresses {
     swarm: Vec<Multiaddr>,
+    api: SocketAddr,
```
koivunej, line 278: I looked over an older go-ipfs config file which only seems to support a single API listening address *in multiaddr*. If this hasn't changed I'd rename all variables touching this `Addresses.api` as singular from plurals.

http/src/config.rs
```diff
@@ -52,27 +84,24 @@ pub fn initialize(
...
-    profiles: Vec<String>,
+    profiles: Vec<Profile>,
 ) -> Result<(), InitializationError> {
     use multibase::Base::Base64Pad;
     use prost::Message;
     use std::io::BufWriter;
 
+    let api_addrs = match profiles[0] {
+        Profile::Test => "127.0.0.1:0",
+        Profile::Default => "127.0.0.1:4004",
```
koivunej, line 95: I think string literals aren't needed here `([127, 0, 0, 1], 0)` could probably be used, or at least could be parsed following this `match` or on the match arms? Or a [`multiaddr!`](https://docs.rs/parity-multiaddr/0.9.3/parity_multiaddr/macro.multiaddr.html) macro could produce the required multiaddr, not sure which way it should be done to have it the easiest.

http/src/config.rs
```diff
@@ -167,7 +197,7 @@ pub fn load(config: File) -> Result<(ipfs::Keypair, Vec<Multiaddr>), LoadingErro
         });
     }
 
-    Ok((kp, addresses.swarm))
+    Ok((kp, addresses.swarm, addresses.api))
```
koivunej, line 200: `CompatibleConfigFile` is probably currently private and that is the reason behind using tuples here along I guess? Tuples probably do not scale up to adding the bootstrap addresses so we'll need to remember to change this then at last.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+* fix(http): create `Profile` abstraction [#421]
```
koivunej, line 3: I would rather call this a feat(ure)

http/src/config.rs
niklaslong: I initially started out with a `MultiAddr` though switched to parsing to a `SocketAddr` as it was the simplest to use with warp; otherwise I needed to implement `Into<SocketAddr>` for `Protocol` if I remember correctly. Would this be preferred?

http/src/config.rs
koivunej: Aah apologies, read the original comment wrong, you were right on the empty config file which I misread as empty `.Addresses.API` :)

http/src/config.rs
niklaslong: Actually the go implementation uses `MultiAddr` as well as a capitalised `API` key:
`"API": "/ip4/127.0.0.1/tcp/5001"

http/src/main.rs
```diff
@@ -185,17 +190,29 @@ fn main() {
...
 
-    warp::serve(routes).bind_with_graceful_shutdown(([127, 0, 0, 1], 0), async move {
+    let components = listening_addr.iter().collect::<Vec<_>>();
+
+    use std::net::SocketAddr;
+    let socket_addr =
+        if let (Protocol::Ip4(ip), Protocol::Tcp(port)) = (&components[0], &components[1]) {
+            SocketAddr::new(ip.clone().into(), *port)
+        } else {
+            panic!("Couldn't convert MultiAddr into SocketAddr")
```
koivunej, line 212:
```suggestion
            panic!("Couldn't convert MultiAddr into SocketAddr: {}", listening_addr)
```

http/src/main.rs
```diff
@@ -185,17 +190,29 @@ fn main() {
...
 
     let routes = v0::routes(ipfs, shutdown_tx);
     let routes = routes.with(warp::log(env!("CARGO_PKG_NAME")));
 
     let ipfs = ipfs.clone();
 
-    warp::serve(routes).bind_with_graceful_shutdown(([127, 0, 0, 1], 0), async move {
+    let components = listening_addr.iter().collect::<Vec<_>>();
+
+    use std::net::SocketAddr;
```
koivunej, line 207: I wonder if this could be moved to the beginning of the function, like the rest of the `use`s?

http/src/main.rs
```diff
@@ -185,17 +190,29 @@ fn main() {
...
+    let socket_addr =
+        if let (Protocol::Ip4(ip), Protocol::Tcp(port)) = (&components[0], &components[1]) {
```
koivunej, line 209: I think this needs an additional check that the length of components is 2... Now, how to realize that.

My simplest solution (which doesnt turn out simple at all) is always to "just" call iter.next() three times, but that gets complicated quite fast in order to honor the "cannot call `iter.next()` after it has returned None" agreement.

Perhaps [slice pattern matching](https://doc.rust-lang.org/edition-guide/rust-2018/slice-patterns.html) could work here together with the vec collecting?

koivunej (commented):
Still a few minor issues, otherwise this is looking very much ready to go.

http/src/main.rs
koivunej: Alternative would be to not `use` but to have the full path on the single use.

koivunej (approved):
looking good

bors r+

PR #422, opened 2020-10-27
Further reduction of ipfs-http compilation times

lomereiter:
See https://github.com/seanmonstar/warp/issues/619; YMMV, I observed 30-40% speedup on my machine.

As hinted [here](https://rust-classes.com/chapter_6_3.html), a more insightful way to compare is to run
```
cargo +nightly rustc --bin ipfs-http -- -Z time-passes
```
and look at the `monomorphization_collector_graph_walk` entry - this particular pass executes ~3x faster with this change.


koivunej (approved):
Thank you very much! For me incremental builds go from 44s (master) to 22s (this PR) when jumping between branches using lld-9. Without lld-9 incremental builds are still faster: 52s vs. 30s.

bors r+

koivunej:
bors ping

PR #423, opened 2020-10-29
http: configuration refactor

niklaslong:
This is a wip but opening as a draft to keep track of this follow up to #421. This is also part of #402.


http/src/config.rs
```diff
@@ -146,10 +130,28 @@ fn create(
...
+    let config_path = ipfs_path.join("config");
+
+    let config_file = fs::create_dir_all(&ipfs_path)
+        .map_err(InitializationError::DirectoryCreationFailed)
+        .and_then(|_| {
+            fs::File::create(&config_path).map_err(InitializationError::ConfigCreationFailed)
+        })?;
+
+    serde_json::to_writer_pretty(BufWriter::new(config_file), &config_contents)
         .map_err(InitializationError::ConfigWritingFailed)?;
```
koivunej, line 143: Probably an older bug: since we are using `BufWriter` and `serde_json::to_writer_pretty` does not explicitly say it'll flush the writer (std::io::Write::flush), perhaps we should pass the writer as `&mut` and take care of this ourselves? This will otherwise be done in `drop` and the error would be lost.

http/src/config.rs
```diff
@@ -146,10 +130,28 @@ fn create(
         },
     };
 
-    serde_json::to_writer_pretty(BufWriter::new(config), &config_contents)
+    let config_path = ipfs_path.join("config");
+
+    let config_file = fs::create_dir_all(&ipfs_path)
+        .map_err(InitializationError::DirectoryCreationFailed)
+        .and_then(|_| {
+            fs::File::create(&config_path).map_err(InitializationError::ConfigCreationFailed)
```
koivunej, line 138: Probably an older bug: Will [`std::fs::File::create`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.create) truncate the target? Yes. We should use [`OpenOptions::create_new`](https://doc.rust-lang.org/stable/std/fs/struct.OpenOptions.html#method.create_new) to avoid truncation. The only way the file exists at this point if it was created between src/main.rs checking `!file.exists()` and this create, but we still shouldn't truncate it. The `!file.exists()` from src/main.rs can probably stay, it's a nice sanity check before rsa keygen.

http/src/main.rs
```diff
@@ -73,20 +73,14 @@ fn main() {
                 std::process::exit(1);
             }
 
-            let result = config::initialize(&home, bits, profile);
+            let result = config::init(&home, bits, profile);
 
             match result {
-                Ok(_) => {
-                    let (kp, _, _) = std::fs::File::open(config_path)
```
koivunej, line 80: What's the rationale for *not* reopening?

If I remember correctly I added this as sanity check for the roundtrip working. This would be the `LoadingError::PeerIdMismatch`.

http/src/main.rs
```diff
@@ -158,11 +152,13 @@ fn main() {
...
         let api_link_file = home.join("api");
 
-        let (addr, server) = serve(&ipfs, api_listening_addr);
+        let (addr, server) = serve(&ipfs, config.api_addr);
 
         // shutdown future will handle signalling the exit
         drop(ipfs);
 
+        // We can't simply reuse the address from the config as the test profile uses ephemeral
+        // ports.
```
koivunej, line 161: Good point ... This answers my question I was about to ask :)

koivunej (commented):
Looking good. Found some older bugs we should at least be aware of. One question.

http/src/main.rs
koivunej: After re-reading this a few times it seems that I must have only used some println logging for confirming the peerid before writing and after re-opening, ok to change this as it never asserted anything else than written file can be loaded.

http/src/config.rs
```diff
@@ -146,10 +131,36 @@ fn create(
...
+                .map_err(InitializationError::ConfigCreationFailed)
+        })?;
+
+    let mut writer = BufWriter::new(config_file);
+
+    serde_json::to_writer_pretty(&mut writer, &config_contents)
         .map_err(InitializationError::ConfigWritingFailed)?;
 
-    Ok(())
+    writer.flush().unwrap();
```
koivunej, line 151:
```suggestion
    writer.flush().map_err(InitializationError::ConfigCreationFailed)?;
```

Not too critical though, but as the the errors have been used around here already, it'd be inline with other functionality. The `ConfigCreationFailed` isn't a really good match for this of course.. Another way would be to change the type of `ConfigWritingFailed` to be `Box<dyn std::error::Error + 'static>` and it'd probably work just as fine.

koivunej (approved):
The unwrap remains which I think we should get rid of, by using the std::io::Error wrapping variant or widening the wrapped type in the more fitting from `serde_json::Error` to `Box<dyn std::error::Error + 'static>`.

Good to go after that.

bors d+

http/src/config.rs
niklaslong: Ah yes, was wondering about this, thanks 👍

PR #427, opened 2020-11-05
Further reduction of ipfs-http compilation times, take two

koivunej:
This PR supercedes #422 as bors couldn't handle the "PR is from removed repository" case, so this is made from a branch which has only the same commit as the previous PR. PR uses the balanced tree approach from warp issue for further compilation time speedups. The commit hash differs because I apparently applied it per githubs instructions on top of different commit than the original seems to be, with different metadata.

Closes #422.


aphelionz (approved):
bors r+

PR #428, opened 2020-11-09
Update to go-ipfs 0.7.0 (used in interop tests) 

niklaslong:
We were using `0.6.0`, this upgrades to `0.7.0`. 

I also removed the `--bits` option as `0.7` uses `ed25519` by default.


koivunej (approved):
Looking great!

bors r+

PR #429, opened 2020-11-10
feat: repo locking with fs2

### LINKED ISSUE #243: Repo locking
Not providing any repo locking causes difficult to debug issues as I've hit earlier myself and now others have as well, see #241.

Simplest way could be to fail at [Repo::init](https://github.com/rs-ipfs/rust-ipfs/blob/master/src/repo/mod.rs#L200) time if a lock file exists. If the lock file doesn't exist, one must be created and again removed at exit time. Probably ok to rely on deletion at Drop, require manual recovery (file deletion) unless a (go/js) compatible recovery strategy exists.

While we are not aiming to be 100% repo compatible with the other implementations being compatible here sounds like a solid idea. go-ipfs 0.5 seems to use a simple protocol of creating an empty file `$IPFS_PATH/repo_lock` with the following `/proc/fdinfo/$fd`:

```
$ cat /proc/1159/fdinfo/5
pos:    0
flags:  02100002
mnt_id: 29
lock:   1: POSIX  ADVISORY  WRITE 1159 fd:01:2101410 0 EOF
```

_Originally posted by @koivunej in https://github.com/rs-ipfs/rust-ipfs/issues/241#issuecomment-658802088_

niklaslong:
supersedes #426 and resolves #243. 

This implements repo locking with fs2. The only concern for this crate is that it is poorly maintained but the functionality we need seems stable enough for now. 

Certain tests are currently failing, notably because multiple nodes are being created with the same repo (dag tests for instance). Not sure what the solution to this will be...

Update: running tests with `--test-threads=1` passes all unit tests. The integration tests fail (I think because multiple nodes are being spun up with the same repo). 

Update 2: all green, making sure test nodes use a different temp dir for the repo fixed the tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,7 +56,7 @@ use tracing_futures::Instrument;
 use std::{
     borrow::Borrow,
     collections::{HashMap, HashSet},
-    fmt,
+    env, fmt,
     future::Future,
     ops::{Deref, DerefMut, Range},
     path::PathBuf,
@@ -104,6 +104,7 @@ pub struct Types;
 impl RepoTypes for Types {
     type TBlockStore = repo::fs::FsBlockStore;
     type TDataStore = repo::fs::FsDataStore;
+    type TLock = repo::fs::FsLock;
 }
 
 /// In-memory testing configuration used in tests.
@@ -112,6 +113,7 @@ pub struct TestTypes;
 impl RepoTypes for TestTypes {
     type TBlockStore = repo::mem::MemBlockStore;
     type TDataStore = repo::mem::MemDataStore;
+    type TLock = repo::mem::MemLock;
 }
 
 /// Ipfs node options used to configure the node to be created with [`UninitializedIpfs`].
@@ -177,12 +179,8 @@ impl IpfsOptions {
     ///
     /// Also used from examples.
     pub fn inmemory_with_generated_keys() -> Self {
-        use tempfile::TempDir;
-
-        let tempdir = TempDir::new().expect("tempdir creation failed").into_path();
-
         Self {
-            ipfs_path: tempdir,
+            ipfs_path: env::temp_dir(),
             keypair: Keypair::generate_ed25519(),
             mdns: Default::default(),
             bootstrap: Default::default(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,7 +56,7 @@ use tracing_futures::Instrument;
 use std::{
     borrow::Borrow,
     collections::{HashMap, HashSet},
-    env, fmt,
+    fmt,
     future::Future,
     ops::{Deref, DerefMut, Range},
     path::PathBuf,
@@ -177,8 +177,12 @@ impl IpfsOptions {
     ///
     /// Also used from examples.
     pub fn inmemory_with_generated_keys() -> Self {
+        use tempfile::TempDir;
+
+        let tempdir = TempDir::new().expect("tempdir creation failed").into_path();
+
         Self {
-            ipfs_path: env::temp_dir(),
+            ipfs_path: tempdir,
             keypair: Keypair::generate_ed25519(),
             mdns: Default::default(),
             bootstrap: Default::default(),
--- a/src/repo/fs.rs
+++ b/src/repo/fs.rs
@@ -4,6 +4,7 @@
 
 use crate::error::Error;
 use async_trait::async_trait;
+use std::fs::File;
 use std::path::PathBuf;
 use std::sync::{atomic::AtomicU64, Arc};
 use tokio::sync::Semaphore;
@@ -88,9 +89,6 @@ impl DataStore for FsDataStore {
     }
 }
 
-use fs2::FileExt;
-use std::fs::File;
-
 #[derive(Debug)]
 pub struct FsLock {
     file: Option<File>,
@@ -114,6 +112,7 @@ impl Lock for FsLock {
     }
 
     fn try_exclusive(&mut self) -> Result<(), LockError> {
+        use fs2::FileExt;
         use std::fs::OpenOptions;
 
         let file = OpenOptions::new()
--- a/src/repo/fs.rs
+++ b/src/repo/fs.rs
@@ -8,7 +8,7 @@ use std::path::PathBuf;
 use std::sync::{atomic::AtomicU64, Arc};
 use tokio::sync::Semaphore;
 
-use super::{BlockRm, BlockRmError, Column, DataStore, Lock, RepoCid};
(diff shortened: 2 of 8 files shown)


src/lib.rs
```diff
@@ -177,8 +177,12 @@ impl IpfsOptions {
     ///
     /// Also used from examples.
     pub fn inmemory_with_generated_keys() -> Self {
+        use tempfile::TempDir;
+
+        let tempdir = TempDir::new().expect("tempdir creation failed").into_path();
+
         Self {
-            ipfs_path: env::temp_dir(),
+            ipfs_path: tempdir,
```
koivunej, line 185: does this mean that the dev system gets polluted with temporary directories?

src/repo/mod.rs
```diff
@@ -235,8 +238,34 @@ impl TryFrom<RequestKind> for RepoEvent {
...
+        use std::fs::OpenOptions;
+
+        let file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(create)
+            .open(path)
+            .unwrap();
+
+        FileExt::try_lock_exclusive(&file)?;
```
koivunej, line 257: I guess this way of invocation is ok as well, perhaps more usual way to call would be  `file.try_lock_exclusive()?`.

src/repo/mod.rs
```diff
@@ -235,8 +238,34 @@ impl TryFrom<RequestKind> for RepoEvent {
...
+
+        Ok(Lock { file })
+    }
+}
+
 impl<TRepoTypes: RepoTypes> Repo<TRepoTypes> {
     pub fn new(options: RepoOptions) -> (Self, Receiver<RepoEvent>) {
+        let lockfile_path = options.path.join("repo_lock");
+        let create = !lockfile_path.is_file();
```
koivunej, line 266: I think this will be a TOCTOU issue, which could be worked around by looping on creating and opening for writing but ... I am not entirely sure. Perhaps this is ok for initial attempt.

src/lib.rs
niklaslong: This should only be using when testing and running the examples, as far as I understand, the directory is deleted when the temp dir goes out of scope. It could be explicitly done with `TempDir::close` though not sure where that would have to go `repo::shutdown` perhaps? I believe tests rely on drop at the moment (no explicit call to shutdown afaik), so this would need changing elsewhere as well. I'll investigate 👍

src/repo/mod.rs
niklaslong: We could actually remove this check entirely as `OpenOptions::create` will create if it doesn't exist and simply open if it does. I *think* that would remove the TOCTOU issue?

koivunej: Yeah I think it will take care of this.

I was initially thinking that a loop of "create new" and "open existing for write" was needed but can't see how it would be better than just a single "create if not exist" file open. The important thing is not to ever create a new file to replace the old, which I think is not possible with plain open, it'd need remove as well so that's not something done by mistake :)

src/lib.rs
niklaslong: You were 💯 on the money there, so I tried something else, not sure if it is the right approach but it seems to work. This only creates a lock if the repo is on disk, the lock struct is empty if in memory (which is what we use for tests). I've checked that this implementation doesn't leave any temp dirs.

src/repo/mod.rs
```diff
@@ -116,6 +117,13 @@ pub trait DataStore: PinStore + Debug + Send + Sync + Unpin + 'static {
     async fn wipe(&self);
 }
 
+#[async_trait]
```
koivunej, line 120: This seems unnecessary, or is there a reason why this needs to be an `#[async_trait]` as well which is not obvious?

src/repo/mem.rs
```diff
@@ -648,6 +648,17 @@ pub enum PinUpdateError {
     CannotUnpinDirectOnRecursivelyPinned,
 }
 
+use std::path::Path;
+
+#[derive(Debug)]
+pub struct MemLock;
```
koivunej, line 654: could document that this is used for the inmemory repos, no need to do anything

src/repo/fs.rs
```diff
@@ -88,5 +88,51 @@ impl DataStore for FsDataStore {
...
+impl Lock for FsLock {
+    fn new(path: &Path) -> std::io::Result<Self> {
+        use std::fs::OpenOptions;
+
+        let file = OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(true)
+            .open(path)
+            .unwrap();
```
koivunej, line 109: In general, under which conditions does this panic? The home is a read-only directory?

While in general this panicing for such an initialization error would be ok for *an application*, perhaps we should find a way to return the error, and only do this at the `Repo::init` time? That would probably mean taking in a PathBuf instead, holding on to it until we can open it as a file, adding an error type which would only be used internally (or be public), cast to `anyhow::Error` on the `UninitializedIpfs::start` https://github.com/rs-ipfs/rust-ipfs/blob/d3c84ee52e926c1c2370f09515db02d5d9dd584c/src/lib.rs#L335

koivunej (commented):
Minor nit on the perhaps extra `#[async_trait]`. Bigger issue with the unwraps. Great that this started working.

As we've already discussed, the `fs2` dependency while not actively maintained is probably ok for the short/medium-term. I can't see many reasons why this functionality would have to change though, given the fact that libc stuff doesn't change to often.

src/repo/mod.rs
niklaslong: Yep, it's not needed 👍

src/repo/fs.rs
niklaslong: Jotting down some thoughts here: 

> In general, under which conditions does this panic? The home is a read-only directory?

Bad path, permissions, mentioned [here](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#errors). This specific error path might also be relevant for the rest of the API (at least the `fs` impl). 

`try_lock_exclusive` errors are os errors, as an example this is [the code for unix](https://github.com/danburkert/fs2-rs/blob/master/src/unix.rs#L46-L48). We could also go the rustc route and block whilst waiting to acquire a lock though I'm thinking a nice `Another IPFS node is already running with this peer id` would be better.`

In order to lock the repo on `init` instead of `new`, I'm thinking the `Lock` struct will need a state key with an Enum variant (i.e. Lock with pathbuf in `repo::new`, actually create the Lock in `init`). Something like [this](https://github.com/rust-lang/cargo/blob/02205f5403d33d5d1666fc7b9f6fd321de511878/src/cargo/util/flock.rs#L13-L25)?

koivunej (approved):
This one is looking good to go!

We did cover a lot of ground in chat outside of this higher latency review interface.

bors r+

PR #431, opened 2020-11-17
Update libp2p and tokio

ljedrz:
A mostly straightforward update to more recent versions of these dependencies.

Filing it as a draft due to the `block_exchange::long_get_block` mysteriously failing, at least locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,7 +40,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "63dd91889c49327ad7ef3b500fd1109dbd3c509a03db0d4a9ce413b79f575cb6"
 dependencies = [
  "block-cipher",
- "byteorder 1.3.4",
+ "byteorder",
  "opaque-debug 0.3.0",
 ]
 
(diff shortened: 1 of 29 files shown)


koivunej:
Looked into getting `ipfs-http` to run with the help of [`tokio_compat_02`](https://docs.rs/tokio-compat-02/0.1) in https://github.com/ljedrz/rust-ipfs/pull/1. The way I understand the state it'll end up to is:

1. tokio-0.3 running as configured in the `http/src/main.rs`
2. tokio-0.2 running as single threaded from `tokio_compat_02`
3. all warp work being executed on tokio-0.2
4. therefore all `ipfs` work (except for background task, swarm spawns) will happen on single threaded tokio-0.2
5. because single threaded tokio-0.2, some `ipfs` work will fail, like FsStore using **`spawn_blocking` on tokio-0.3**
6. `spawn_blocking` on tokio-0.3 will fail because of there is no tokio-0.3 runtime available on the single tokio-0.2 thread

(emphasis on `spawn_blocking` on tokio-0.3 because this threw me off a bit, but makes sense: nowhere in the tokio-0.2 path is the tokio-0.3 reactor made available locally to thread, so it will not find it.)

I think we can land this only when https://github.com/seanmonstar/warp/pull/741 is done along with the `bytes` 0.6 upgrades happening, which may or may not be linked on the warp PR.

ljedrz:
Further `tokio 0.3` work in `warp`: https://github.com/seanmonstar/warp/pull/753.

koivunej:
Status of the upgrade seems that we are still awaiting for the warp PR to go through. At the same time there are at least two major PRs ongoing in rust-libp2p 1907 and the larger 1887. Looks like this will be a big one since we need to adapt to the fixed size multihash as well, unless the two work side by side.

koivunej:
Rebased, superceded in #446.

PR #434, opened 2020-12-23

fetchadd:

### DIFF
--- a/src/repo/kv.rs
+++ b/src/repo/kv.rs
@@ -11,12 +11,12 @@ use std::str::{self, FromStr};
 use tracing_futures::Instrument;
 
 #[derive(Debug)]
-pub struct KvDbStore {
+pub struct KvDataStore {
     path: PathBuf,
     db: Option<Db>,
 }
 
-impl KvDbStore {
+impl KvDataStore {
     fn _put(&self, key: &str, value: &str) -> Result<(), Error> {
         let db = self.get_db();
 
@@ -46,9 +46,9 @@ impl KvDbStore {
 }
 
 #[async_trait]
-impl DataStore for KvDbStore {
-    fn new(root: PathBuf) -> KvDbStore {
-        KvDbStore {
+impl DataStore for KvDataStore {
+    fn new(root: PathBuf) -> KvDataStore {
+        KvDataStore {
             path: root,
             db: None,
         }
@@ -58,8 +58,8 @@ impl DataStore for KvDbStore {
         let db = sled::open(self.path.as_path())?;
 
         unsafe {
-            let kv_ref = self as *const KvDbStore;
-            let kv_mut = kv_ref as *mut KvDbStore;
+            let kv_ref = self as *const KvDataStore;
+            let kv_mut = kv_ref as *mut KvDataStore;
             (*kv_mut).db = Some(db);
         }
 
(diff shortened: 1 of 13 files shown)


src/repo/kv.rs
```diff
@@ -0,0 +1,373 @@
...
+        let db = config
+            .mode(DbMode::HighThroughput)
+            .path(self.path.as_path())
+            .open()?;
+
+        unsafe {
+            let kv_ref = self as *const KvDataStore;
+            let kv_mut = kv_ref as *mut KvDataStore;
+            (*kv_mut).db = Some(db);
+        }
```
koivunej, line 69: Took me a while to decipher this but yeah ... `init(&self)` is an unfortunate wrinkle in the current design. Thanks for not changing this, this is much easier to review at least.. I cannot really see any workarounds for this at the moment even if this is quite unsafe.. Did you get any segfaults around this when testing? I wonder if there should at least be a memory barrier.

koivunej (commented):
I think this is looking good. There's one comment regarding the unfortunate unsafe block but thanks a lot for doing this!

Using `sled` more in the future is something we'll want to heavily consider.

src/repo/kv.rs
fetchadd: Yes, `init(&self)` brings a trouble. I got no segfaults when testing.

koivunej:
Did you try out the `pinstore_interface_tests!`, as in add this:

```diff
diff --git a/src/repo/kv.rs b/src/repo/kv.rs
index 19f0149..213b592 100644
--- a/src/repo/kv.rs
+++ b/src/repo/kv.rs
@@ -371,3 +371,6 @@ fn is_pinned(db: &KvDataStore, block: &Cid) -> Result<bool, Error> {
         Err(e) => Err(e),
     }
 }
+
+#[cfg(test)]
+crate::pinstore_interface_tests!(common_tests, crate::repo::kv::KvDataStore::new);
```

The macro will expand to a few test fns which aim to test the api behaviour of the store implementations, defined in [common_tests.rs](https://github.com/rs-ipfs/rust-ipfs/blob/master/src/repo/common_tests.rs).

There are five successes and three failures:

```
test repo::kv::common_tests::cannot_recursively_unpin_unpinned ... FAILED
test repo::kv::common_tests::pin_direct_twice_is_good ... ok
test repo::kv::common_tests::indirect_can_be_pinned_directly ... ok
test repo::kv::common_tests::pin_recursive_pins_all_blocks ... ok
test repo::kv::common_tests::cannot_unpin_indirect ... FAILED
test repo::kv::common_tests::cannot_pin_recursively_pinned_directly ... FAILED
test repo::kv::common_tests::direct_and_indirect_when_parent_unpinned ... ok
test repo::kv::common_tests::can_pin_direct_as_recursive ... ok
```

I see there aren't too many comments in those common tests so please ask away if they are difficult to decipher. Could you take a look and see if the failures are quick to fix?

Overall I am thinking that the three failures might not be critical enough to block the PR as we will soon have largeish changes as we'll be able to update to tokio 1.0, so I'd rather get this PR in before those happen.

src/repo/kv.rs
koivunej: It might be good enough for now to use mutex and clone the Db as in https://github.com/fetchadd/rust-ipfs/pull/1

fetchadd:
I am going to try to fix the failures and apply the changes to the newest master with no verbose commits..

fetchadd:
Sorry for I don't known to how squash commits away from the branch in the PR, so I reopen it in https://github.com/rs-ipfs/rust-ipfs/pull/438 @koivunej

koivunej:
> Sorry for I don't known to how squash commits away from the branch in the PR, so I reopen it in #439

This is fine! For future reference, I usually do the git trickery with [interactive rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).

PR #438, opened 2021-01-04

fetchadd:


fetchadd:
The PR is the reopen of [434](https://github.com/rs-ipfs/rust-ipfs/pull/434), for I don't know how to reopen a PR directly. 

The changes is applied to the newest master,  and adding new `pinstore_interface_tests` for the KvDataStore.

PR #439, opened 2021-01-05

fetchadd:

This PR is the reopen of [PR434](https://github.com/rs-ipfs/rust-ipfs/pull/434). For there is some verbose commits like "replaceing leveldb with sled",  the master branch of fetchadd/rust-ipfs in the PR434 is deprecated. Now the same changes are applied to the staging branch which is synced with the master of rust-ipfs/rust-ipfs. So Thanks for the PR of fetchadd/rust-ipfs from @koivunej which uses mutex to work around init(&self) and unsafe, but as the master is deprecated, I can only copy the code to staging branch. 

Accorrding to the review from @ koivunej in [PR434](https://github.com/rs-ipfs/rust-ipfs/pull/434), the `pinstore_interface_tests` is added to KvDataStore, and now all tests are passed.

Closes #434.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,7 +103,7 @@ impl<T: RepoTypes> IpfsTypes for T {}
 pub struct Types;
 impl RepoTypes for Types {
     type TBlockStore = repo::fs::FsBlockStore;
-    type TDataStore = repo::fs::FsDataStore;
+    type TDataStore = repo::kv::KvDataStore;
     type TLock = repo::fs::FsLock;
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,7 +103,10 @@ impl<T: RepoTypes> IpfsTypes for T {}
 pub struct Types;
 impl RepoTypes for Types {
     type TBlockStore = repo::fs::FsBlockStore;
+    #[cfg(feature = "sled_data_store")]
     type TDataStore = repo::kv::KvDataStore;
+    #[cfg(not(feature = "sled_data_store"))]
+    type TDataStore = repo::fs::FsDataStore;
     type TLock = repo::fs::FsLock;
 }
 
--- a/src/repo/kv.rs
+++ b/src/repo/kv.rs
@@ -1,0 +1,387 @@
+use super::{Column, DataStore};
+use crate::error::Error;
+use crate::repo::{PinKind, PinMode, PinStore, References};
+use async_trait::async_trait;
+use cid::{self, Cid};
+use futures::stream::{StreamExt, TryStreamExt};
+use sled::{self, Config as DbConfig, Db, Mode as DbMode};
+use std::convert::Into;
+use std::path::PathBuf;
+use std::str::{self, FromStr};
+use std::sync::Mutex;
+use tracing_futures::Instrument;
+
+#[derive(Debug)]
+pub struct KvDataStore {
+    path: PathBuf,
+    db: Mutex<Option<Db>>,
+}
+
+impl KvDataStore {
+    fn _put(&self, key: &str, value: &str) -> Result<(), Error> {
+        let db = self.get_db();
+
+        let _ = db.insert(key, value)?;
+
+        Ok(())
+    }
+
+    fn _remove(&self, key: &str) -> Result<(), Error> {
+        let db = self.get_db();
+
+        match db.remove(key) {
+            Ok(_) => Ok(()),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    fn _apply_batch(&self, batch: sled::Batch) -> Result<(), Error> {
+        let db = self.get_db();
+
+        Ok(db.apply_batch(batch)?)
+    }
+
+    fn get_db(&self) -> Db {
+        self.db.lock().unwrap().as_ref().unwrap().clone()
+    }
+}
+
+#[async_trait]
+impl DataStore for KvDataStore {
+    fn new(root: PathBuf) -> KvDataStore {
+        KvDataStore {
+            path: root,
+            db: Default::default(),
+        }
+    }
+
+    async fn init(&self) -> Result<(), Error> {
+        let config = DbConfig::new();
+
+        let db = config
+            .mode(DbMode::HighThroughput)
+            .path(self.path.as_path())
+            .open()?;
+
+        let mut g = self.db.lock().unwrap();
+        assert!(g.is_none());
+        *g = Some(db);
+
+        Ok(())
+    }
+
+    async fn open(&self) -> Result<(), Error> {
+        Ok(())
+    }
+
+    /// Checks if a key is present in the datastore.
+    async fn contains(&self, _col: Column, _key: &[u8]) -> Result<bool, Error> {
+        Err(anyhow::anyhow!("not implemented"))
+    }
+
+    /// Returns the value associated with a key from the datastore.
+    async fn get(&self, _col: Column, _key: &[u8]) -> Result<Option<Vec<u8>>, Error> {
+        Err(anyhow::anyhow!("not implemented"))
+    }
+
+    /// Puts the value under the key in the datastore.
+    async fn put(&self, _col: Column, _key: &[u8], _value: &[u8]) -> Result<(), Error> {
+        Err(anyhow::anyhow!("not implemented"))
+    }
+
+    /// Removes a key-value pair from the datastore.
+    async fn remove(&self, _col: Column, _key: &[u8]) -> Result<(), Error> {
+        Err(anyhow::anyhow!("not implemented"))
+    }
+
+    /// Wipes the datastore.
+    async fn wipe(&self) {
+        todo!()
+    }
+}
+
+#[async_trait]
+impl PinStore for KvDataStore {
+    async fn is_pinned(&self, block: &Cid) -> Result<bool, Error> {
+        is_pinned(self, block)
+    }
+
+    async fn insert_direct_pin(&self, target: &Cid) -> Result<(), Error> {
+        let already_pinned = get_pinned_mode(self, target)?;
+
+        let mut batch = sled::Batch::default();
+
+        match already_pinned {
+            Some(PinMode::Direct) => return Ok(()),
+            Some(PinMode::Recursive) => return Err(anyhow::anyhow!("already pinned recursively")),
+            Some(PinMode::Indirect) => {
+                let pin_key = get_pin_key(target, &PinMode::Indirect);
+                batch.remove(pin_key.as_str());
+            }
+            _ => {}
+        }
+
+        let direct_key = get_pin_key(target, &PinMode::Direct);
+
+        batch.insert(direct_key.as_str(), "");
+
+        Ok(self._apply_batch(batch)?)
+    }
+
+    async fn insert_recursive_pin(
+        &self,
+        target: &Cid,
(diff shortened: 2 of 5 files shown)


koivunej (approved):
This is looking great, thank you. I might be adding a few stylistic tweaks, changelog and comments and will be cc'ing you in a follow-up PR.

Closes #434.

Awaiting a sec @ljedrz wants to review as well..

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+use std::convert::Into;
+use std::path::PathBuf;
+use std::str::{self, FromStr};
+use std::sync::Mutex;
+use tracing_futures::Instrument;
+
+#[derive(Debug)]
+pub struct KvDataStore {
+    path: PathBuf,
+    db: Mutex<Option<Db>>,
```
ljedrz, line 17: since the operations on `Db` take `self` by reference, it would probably be best to instead use something like an `OnceCell<Db>` here instead

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    async fn init(&self) -> Result<(), Error> {
+        let config = DbConfig::new();
+
+        let db = config
+            .mode(DbMode::HighThroughput)
+            .path(self.path.as_path())
+            .open()?;
```
ljedrz, line 64: since there is a `DataStore::open`, maybe that one should perform this action instead?

koivunej: I think the `open` needs to be thrown away and I don't think they are even called, hopefully even along with `init` and go full RAII with these components. They are mostly useless after being new'd without some interior mutability going on.

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        Ok(self._apply_batch(batch)?)
+    }
+
+    async fn insert_recursive_pin(
+        &self,
+        target: &Cid,
+        referenced: References<'_>,
+    ) -> Result<(), Error> {
+        let set = referenced
+            .try_collect::<std::collections::BTreeSet<_>>()
```
ljedrz, line 137: nit: since it's used more often, I'd prefer this to be imported

src/repo/kv.rs
koivunej: Is there a stable SyncOnceCell? I think those are still unstable.

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        let mut batch = sled::Batch::default();
+
+        match already_pinned {
+            Some(PinMode::Direct) => return Ok(()),
+            Some(PinMode::Recursive) => return Err(anyhow::anyhow!("already pinned recursively")),
+            Some(PinMode::Indirect) => {
+                let pin_key = get_pin_key(target, &PinMode::Indirect);
+                batch.remove(pin_key.as_str());
+            }
+            _ => {}
```
ljedrz, line 121: I'd prefer this to branch to match against `None` if that's the only other possibility

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        let mut batch = sled::Batch::default();
+        let already_pinned = get_pinned_mode(self, target)?;
+
+        match already_pinned {
+            Some(PinMode::Recursive) => return Ok(()),
+            Some(mode @ PinMode::Direct) | Some(mode @ PinMode::Indirect) => {
+                let key = get_pin_key(target, &mode);
+                batch.remove(key.as_str());
+            }
+            _ => {}
```
ljedrz, line 149: same note on an explicit `None` here if that's the only possibility left

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+#[async_trait]
+impl PinStore for KvDataStore {
+    async fn is_pinned(&self, block: &Cid) -> Result<bool, Error> {
+        is_pinned(self, block)
+    }
+
+    async fn insert_direct_pin(&self, target: &Cid) -> Result<(), Error> {
+        let already_pinned = get_pinned_mode(self, target)?;
+
+        let mut batch = sled::Batch::default();
```
ljedrz, line 112: I was thinking if this wasn't potentially costly (in which case conditional creation would be in order), but I've checked and it isn't :+1:.

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+    async fn remove_recursive_pin(
+        &self,
+        target: &Cid,
+        referenced: References<'_>,
+    ) -> Result<(), Error> {
+        let set = referenced
+            .try_collect::<std::collections::BTreeSet<_>>()
+            .await?;
+
+        if is_not_pinned_or_pinned_indirectly(self, target)? {
```
ljedrz, line 190: this potential early return should go before the creation of `set`

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        for cid in &set {
+            let already_pinned = get_pinned_mode(self, cid)?;
+
+            match already_pinned {
+                Some(PinMode::Recursive) | Some(PinMode::Direct) => continue, // this should be unreachable
+                Some(PinMode::Indirect) => {
+                    let indirect_key = get_pin_key(cid, &PinMode::Indirect);
+                    batch.remove(indirect_key.as_str());
+                }
+                _ => {}
```
ljedrz, line 209: same remark on explicit `None` as earlier

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        let min_key = "pin.0.0000000000000000000000000000000000000000000000";
+        assert_eq!(min_key.len(), 52);
+
+        let iter = db.range(min_key..);
+        let mut all_keys: Vec<String> = vec![];
+
+        for item in iter {
+            if item.is_err() {
+                continue;
+            }
```
ljedrz, line 232: `for item in iter.filter(|item| item.is_ok())` would be more concise

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+
+            let (raw_key, _) = item.unwrap();
+            let key = String::from(String::from_utf8_lossy(raw_key.as_ref()));
```
ljedrz, line 235: the outer conversion is redundant

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+                                        }
+                                    }
+                                }
+                                Ok(_) => {}
+                                Err(e) => return Err(e.into()),
+                            }
+                        }
+                    }
+                }
+                Ok(_) => {}
```
ljedrz, line 333:
```suggestion
                Ok(None) => {}
```

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+fn pin_mode_literal(pin_mode: &PinMode) -> &'static str {
+    match pin_mode {
+        PinMode::Direct => "d",
+        PinMode::Indirect => "i",
+        PinMode::Recursive => "r",
+    }
+}
+
+fn get_pin_key(cid: &Cid, pin_mode: &PinMode) -> String {
+    format!("pin.{}.{}", pin_mode_literal(pin_mode), cid.to_string())
```
ljedrz, line 351:
```suggestion
    format!("pin.{}.{}", pin_mode_literal(pin_mode), cid)
```

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+
+fn get_pinned_mode(kv_db: &KvDataStore, block: &Cid) -> Result<Option<PinMode>, Error> {
+    for mode in &[PinMode::Direct, PinMode::Recursive, PinMode::Indirect] {
+        let key = get_pin_key(block, mode);
+
+        let db = kv_db.get_db();
+
+        match db.get(key.as_str()) {
+            Ok(Some(_)) => return Ok(Some(mode.clone())),
+            Ok(_) => {}
```
ljedrz, line 362:
```suggestion
            Ok(None) => {}
```

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+fn get_pinned_mode(kv_db: &KvDataStore, block: &Cid) -> Result<Option<PinMode>, Error> {
+    for mode in &[PinMode::Direct, PinMode::Recursive, PinMode::Indirect] {
+        let key = get_pin_key(block, mode);
+
+        let db = kv_db.get_db();
+
+        match db.get(key.as_str()) {
+            Ok(Some(_)) => return Ok(Some(mode.clone())),
```
ljedrz, line 361: isn't `PinMode` a `Copy` type?

src/repo/kv.rs
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    Ok(None)
+}
+
+fn is_pinned(db: &KvDataStore, block: &Cid) -> Result<bool, Error> {
+    match get_pinned_mode(db, block) {
+        Ok(Some(_)) => Ok(true),
+        Ok(_) => Ok(false),
```
ljedrz, line 373:
```suggestion
        Ok(None) => Ok(false),
```

src/repo/kv.rs
fetchadd: Good advice, the result of `String::from_utf8_lossy(raw_key.as_ref())` (is a Cow<'_, str>) should be converted String at the push stage, but can the result of `.into_owned` is a `String`?

src/lib.rs
```diff
@@ -103,7 +103,7 @@ impl<T: RepoTypes> IpfsTypes for T {}
 pub struct Types;
 impl RepoTypes for Types {
     type TBlockStore = repo::fs::FsBlockStore;
-    type TDataStore = repo::fs::FsDataStore;
+    type TDataStore = repo::kv::KvDataStore;
```
ljedrz, line 106: both options should probably be feature-gated, with one being a default feature

src/repo/kv.rs
ljedrz: what about https://crates.io/crates/once_cell? it's post `1.0`

src/repo/kv.rs
ljedrz: yes: `Cow::into_owned` will create a `String` in this case

koivunej: the filter below can be done before the lossy utf-8 conversion, but I was thinking that perhaps the whole of the str conversion could be perhaps avoided.

src/repo/kv.rs
fetchadd: Maybe it is better for not importing a new crates :)

ljedrz: performing a `lock()` with every database operation (in addition to ones it's already using) is not ideal, though

src/repo/kv.rs
ljedrz: yeah, I was wondering the same; strings are always UTF-8 validated in Rust, and it's not necessarily needed here

src/repo/kv.rs
fetchadd: Got it, I have check once_cell, it is a good choice in this situation 👍 .

ljedrz (approved):
LGTM :+1:

src/repo/kv.rs
koivunej: very much agreed.

koivunej:
Thank you very much @fetchadd!

bors r+

Issue #458, opened 2021-02-06
Release ipfs 0.3, ipfs-unixfs 0.2.1


koivunej:
It's been a while since the last release given all of the dependency churn in the past. While bug/feature-wise there hasn't been much progress, I think a 0.3 could be cut next week.

Tails:
Would this include bumped dependencies? Because I cannot compile rust-ipfs together with rocket-0.5 for example:

````cargo
# cargo.toml
[package]
name = "test_rocket_ipfs_deps"
version = "0.1.0"
edition = "2018"

[dependencies]
# this we want
rocket = "0.5.0-rc.1"

# ... but this is all we can have
# rocket = "0.4.1"

# err: failed to select a version for `if-addrs-sys`
ipfs = { git = "https://github.com/rs-ipfs/rust-ipfs" }

# err: mismatched types (futures impl)
# ipfs = "0.2.1" 

````

````rust
# ./src/main.rs
fn main() {
    println!("Hello, world!");
}
````

koivunej:
Hi and thanks for the report @Tails! I'll try to find some time this week to see if the deps could be bumped easily.

Tails:
By the way great job on this IPFS lib! Very stable, pleasant code style and easy api. Hope you'll get the/a next funding round!

koivunej:
@Tails  I posted a WIP #463. The "multiaddr don't contain peerid" assumption, while nicely documented, has been changed in a libp2p which will take a bit more time to tackle. Other than that... I don't think there's a huge amount of work here. If you want to continue exploration, using that branch directly would be a way forward.

Please be aware of our status and the lack of features, the lack of dht publishing #418 being the most prominent. Also while content can be discovered, but downloading it will be slow.

Tails:
Thanks for the heads-up!

Tails:
Unfortunately I'm still getting the same mismatch, even after cleaning:

```rust
error[E0308]: mismatched types
    | 
   ::: cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.7.0/src/fs/read_dir.rs:16:50
    |
16  |   pub async fn read_dir(path: impl AsRef<Path>) -> io::Result<ReadDir> {
    |                                                    -------------------
    |                                                    |
    |                                                    one of the expected opaque types
    |                                                    one of the found opaque types
...
267 |       pub async fn file_type(&self) -> io::Result<FileType> {
    |                                        --------------------
    |                                        |
[...]

PR #464, opened 2021-07-16
fix(release 0.3.0: rewrite FsBlockStore.list; crate compiles as dependency again

Christian7573:
This PR rewrites `FsBlockStore.list` using for and while let loops, which removes a compilation error when using the `ipfs` crate as a dependency from another project. See issue #458 for relevant discussion.

- [x] **New** code is “linted” i.e. code formatting via rustfmt and language idioms via clippy
- [x] There are no extraneous changes like formatting, line reordering, etc. Keep the patch sizes small!
- [x] There is suitable documentation. In our case, this means:
    - [x] Each command has a usage example and API specification
    - [x] Rustdoc tests are passing on all code-level comments
    - [ ] Differences between Rust’s IPFS implementation and the Go or JS implementations are explained - N/A?
- [x] Additions to CHANGELOG.md files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * chore: update a lot of dependencies including libp2p, tokio, warp [#446]
 * fix: rename spans (part of [#453])
 * fix: connect using DialPeer instead of DialAddress [#454]
+* fix: crate compiles successfully as a dependency again [#464]
 
 [#429]: https://github.com/rs-ipfs/rust-ipfs/pull/429
 [#428]: https://github.com/rs-ipfs/rust-ipfs/pull/428
@@ -19,6 +20,7 @@
 [#446]: https://github.com/rs-ipfs/rust-ipfs/pull/446
 [#453]: https://github.com/rs-ipfs/rust-ipfs/pull/453
 [#454]: https://github.com/rs-ipfs/rust-ipfs/pull/454
+[#464]: https://github.com/rs-ipfs/rust-ipfs/pull/464
 
 # 0.2.1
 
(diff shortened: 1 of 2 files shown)


koivunej:
There are few fmt issues and I guess clippy as well. Sorry about that.

koivunej:
It seems that there's something wrong with vcpkg.. Trying to find time to look into this.

koivunej:
Thanks for this. I don't really understand why this works and the previous didn't work; wonder if there's a rustc bug hidden here we should report. Do you have an understanding you would try to share, other than "not using combinators" clears up errors in combinator code?

I'm thinking that tomorrow I'll PR my #470 into this PR so that we'll have a test case for this, and had another comment in mind, I'll check if it's valid. After my tomorrow-ish PR, this should be ready to be rebased for CI/clippy fixes (#468) and finally merging.

koivunej:
I got closer to the issue in #470 and deduced it was caused by a trouble inferring the two lifetimes generated by `async-trait` seen [here in the expansion of the original code](https://gist.github.com/koivunej/d6abccb4133839eeab8b36992f1a95fa#file-main-rs-L118-L174) (with some tracing parts removed). Moving the method body as `FsBlockStore::list0` outside the `#[async_trait] impl BlockStore for FsBlockStore { ... }` and just calling the new method in the `BlockStore::list` implementation worked around the issue.

I would prefer to keep the current "less buffering" version if possible, even if there's no concurrent execution of the streams. Could you peek at #470 and see if it matches your understanding of the compilation failure?

Christian7573:
#470 works for me. Awesome!

Issue #465, opened 2021-07-26
Absence of time lead to 1.1.1970 in tar archieve which is incompatible with FAT32


Scondo:
**Describe the bug**
When there is no time specified for file in ipfs record tar file produced for directory set time of this file to 1.1.1970 (start of Unix epoch). However extracting such archive to FAT32 leads to error as start of MS-DOS epoch is 1.1.1980 and no file in FAT32 could have time before that.

I would agree that if it's intentionally set file time to 1.1.1970 - this is a problem of FAT32, but if it's just absence of time - this could be any time in fact. So there is no reason to stick to start of Unix epoch- it may be time of archive creation as well. 

**To Reproduce**
Get tar file constructed by https://github.com/rs-ipfs/rust-ipfs/blob/4bce4679de6c4f206864de843f14e2c273560dcb/http/src/v0/root_files.rs#L96 (I found it within very_long_file_and_symlink_names test) and try to extract it to FAT32.

At least crate tar used by rust-ipfs and 7z on windows produce an error.

**Expected behavior**
I don't know for any existing filesystem, may be some use even later date as limitation, so it's definitely not good idea to set default time to 1.1.1980.

Most obvious solution is to set time of file to time of archive creation. Another - to use 1.1.2001 as start of last notable epoch (https://en.wikipedia.org/wiki/Epoch_(computing))

It's not a big deal for now, so I expect discussion.

koivunej:
> I would agree that if it's intentionally set file time to 1.1.1970

I think this was either by looking at the tar archives produced (at then) by go-ipfs or looking at what gnu tar does by default. Or maybe it was a comment on hiding this information which isn't really needed. The only problem with using another timestamp (or a variable timestamp) is that the tar archives become "less canonical" and are even less hashable. This should had been all commented with the code obviously.

While the test (fix by you in #466) was using tempdir to just get the contents of a file, I think an extractor application should handle special cases like FAT32 not supporting unix epochs as datetimes... So I'm actually quite surprised of this:

> At least crate tar used by rust-ipfs and 7z on windows produce an error.

I assume you used a quite recent 7z? For my own curiosity, what version of Windows did you test this on?

This issue as a whole was one of the reasons I didn't want to add any feature to write to filesystem as I couldn't see how could I test it on all systems. This is an important bug report for that front aside from the test fix as well, thanks for that!

Scondo:
>I assume you used a quite recent 7z? 
[...]

PR #466, opened 2021-07-26
In-memory reading from tar archives instead of extract to temp folder

Scondo:
Get rid of work with filesystem in fs-unrelated tests.

No more problems like https://github.com/rs-ipfs/rust-ipfs/issues/465 lead to tests failure.


koivunej:
This does seem like a very obvious fix, thanks for this! I don't really know why I failed to consider an inmemory version here initially. The build is a bit broken but we'll look into getting it fixed this week, so I'll leave this pending until fixed.

koivunej:
Not approving for workflow run. Thanks for small PR, lets see if we can just let bors handle this.

bors r+

Issue #258, opened 2020-07-23
use hash_hasher in hashmaps/sets where the key is a Cid/Multihash


ljedrz:
Since `Cid`s are already cryptographic-hash-based objects, it makes sense not to do too much extra work hashing them again when used as `HashMap`/`HashSet` keys; the [hash_hasher](https://crates.io/crates/hash_hasher) crate could be used to improve the performance.

koivunej:
`hash_hasher` truncates to least significant (last) 8 bytes: https://docs.rs/hash_hasher/2.0.3/src/hash_hasher/lib.rs.html#87-89

I initially thought using seahash would be fast way to fold the hashes with minimal mixing but can't think of a way this would be worse for cryptographic hashes.

ljedrz:
Isn't it what `SipHasher` (the default `HashMap` hasher) essentially does too? In the end you can only have a `u64` key: https://doc.rust-lang.org/src/core/hash/sip.rs.html#258-295.

koivunej:
> Isn't it what

If by it you mean taking the last 8 bytes of a byte slice, I think SipHasher(s) process all of the input and have a multiround mixing, by the looks of it.

If the comparison was to seahash, both seahash and Siphash seem to process all of the input, but seahash has a simpler mixing, I guess.

I think None of the mixing is needed

koivunej:
Adding the tags, this shouldn't be a too difficult refactoring to try. We do not however have a benchmark which could show that one way is better than the other. Perhaps running manually hyperfine against fetch_and_cat example might work?

Perhaps we should have these in some `ipfs::util` for a while:

 - CidHashSet, MultihashHashSet
 - CidHasMap, MultihashMap

If they seem good, we should propose to move them into cid and multihash crates. Though, might not be a good idea to start doing much work against the bounds with these, just try to use them where we now have `Hash(Set|Map)<\S*Cid`. Hmm, seems that we only have those in the src/repo/mem.rs. Oh, well... It should be enough to see if this would be useful though.

PR #467, opened 2021-08-02
Use hash_hasher in hashmaps/sets where the key is a Cid/Multihash #258

CHr15F0x:
This PR implements a fix for #258, ie. using hash_hasher to improve performance in HashMaps/Sets what utilize Cid as key. hash_hasher provides approx. 40% increase in performance in this case.

There are two things that still need to be done:
- code is **not linted** yet, since this will be done in a separate PR by Mirko,
- a simple benchmark is included for the sake of comparing HashedMap<Cid,> vs HashMap<Cid,>; Imho this bench should not be merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ cid = { default-features = false, version = "0.5" }
 trust-dns-resolver = "0.20"
 either = { default-features = false, version = "1.5" }
 futures = { default-features = false, version = "0.3.9", features = ["alloc", "std"] }
-hash_hasher = { default-features = false, version = "2.0.3" } 
+hash_hasher = "2.0.3"
 ipfs-unixfs = { version = "0.2", path = "unixfs" }
 libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-tokio", "mplex", "noise", "ping", "yamux", "dns"], version = "0.34" }
 multibase = { default-features = false, version = "0.8" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ cid = { default-features = false, version = "0.5" }
 trust-dns-resolver = "0.20"
 either = { default-features = false, version = "1.5" }
 futures = { default-features = false, version = "0.3.9", features = ["alloc", "std"] }
+hash_hasher = { default-features = false, version = "2.0.3" } 
 ipfs-unixfs = { version = "0.2", path = "unixfs" }
 libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-tokio", "mplex", "noise", "ping", "yamux", "dns"], version = "0.34" }
 multibase = { default-features = false, version = "0.8" }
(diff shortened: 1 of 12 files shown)


src/subscription.rs
```diff
@@ -156,7 +156,7 @@ impl<T: Debug + Clone + PartialEq, E: Debug + Clone> SubscriptionRegistry<T, E>
                 // subscriptions.lock().unwrap().
                 drop(subscriptions);
 
-                panic!(msg);
+                panic!("{}", msg);
```
koivunej, line 159: I'm thinking this might need to be dropped when rebasing.

Cargo.toml
```diff
@@ -29,6 +29,7 @@ cid = { default-features = false, version = "0.5" }
 trust-dns-resolver = "0.20"
 either = { default-features = false, version = "1.5" }
 futures = { default-features = false, version = "0.3.9", features = ["alloc", "std"] }
+hash_hasher = { default-features = false, version = "2.0.3" }
```
koivunej, line 32:
```suggestion
hash_hasher = "2.0.3"
```

Since I happened to look, the crate defines no features. I wish there was a lint for these.

koivunej (approved):
Looking good!

src/subscription.rs
CHr15F0x: I've just finished the rebase locally and I'm afraid the change remains. Otherwise I get a warning:
```
warning: panic message is not a string literal
   --> src/subscription.rs:160:24
    |
160 |                 panic!(msg);
    |                        ^^^
    |
    = note: `#[warn(non_fmt_panic)]` on by default
    = note: this is no longer accepted in Rust 2021
help: add a "{}" format string to Display the message
    |
160 |                 panic!("{}", msg);
    |                        ^^^^^
help: or use std::panic::panic_any instead
    |
160 |                 std::panic::panic_any(msg);
    |                 ^^^^^^^^^^^^^^^^^^^^^

```

Cargo.toml
CHr15F0x: Many thanks for spotting this, I didn't check if the crate had any features in the first place.

CHr15F0x:
Comparison using `fetch_and_cat` on a 10GiB file shows _some_ slight improvement:
```
Benchmark #1: ./opt-fetch_and_cat QmTR3tCt1SRd96t7rSZpvMPbj48GijPz8crEuuujSygEC1 /ip4/127.0.0.1/tcp/33109/p2p/QmTjjZBvf9fZxUcSqqdXPy135zWwCisGpswAF38jvQ8Y9G > /dev/null
  Time (mean ± σ):     58.706 s ±  0.407 s    [User: 46.545 s, System: 24.116 s]
  Range (min … max):   57.808 s … 59.624 s    100 runs

Benchmark #1: ./fetch_and_cat     QmTR3tCt1SRd96t7rSZpvMPbj48GijPz8crEuuujSygEC1 /ip4/127.0.0.1/tcp/33109/p2p/QmTjjZBvf9fZxUcSqqdXPy135zWwCisGpswAF38jvQ8Y9G > /dev/null
  Time (mean ± σ):     58.838 s ±  0.425 s    [User: 46.751 s, System: 24.169 s]
  Range (min … max):   57.927 s … 59.728 s    100 runs
```

koivunej:
This needs to have the `Cargo.lock` removed, sorry for not seeing that. After that, it's good to go. Thanks for the benchmarking.

bors d+

koivunej:
bors d-

Could you add a CHANGELOG.md entry?

CHr15F0x:
Changelog is there, let's try with Mr b

bors r+

koivunej:
@CHr15F0x could you avoid the merge commit, rebase on master instead?

PR #471, opened 2021-08-16
ci: cache improvements

Mirko-von-Leipzig:
This PR improves the CI caching, roughly halving the time required when caches are hit (no dependency or toolchain changes).

Overview of changes:

- Use [rust-cache action](https://github.com/Swatinem/rust-cache). This forms the bulk of the time improvement as Rust dependencies are now cached.
- Improve npm caching for conformance tests. This trims 2-3 minutes of build time.
- Improve job and step naming.
- Add a manual trigger for the workflow.
- Remove core dump for macos, no longer required.

Further details can be found in the commits.

I've been running into sporadic conformance test failures on windows ([link to run](https://github.com/Mirko-von-Leipzig/rust-ipfs/actions/runs/1135372291)):
```
  1) .pubsub.subscribe
       multiple connected nodes
         should send/receive 100 messages:
     FetchError: request to http://127.0.0.1:50533/api/v0/pubsub/pub?arg=pubsub-tests-C-aIrZS_7OSQMNSVOMszV failed, reason: read ECONNRESET
      at ClientRequest.<anonymous> (node_modules\node-fetch\lib\index.js:1455:11)
      at Socket.socketErrorListener (_http_client.js:475:9)
      at emitErrorNT (internal/streams/destroy.js:106:8)
      at emitErrorCloseNT (internal/streams/destroy.js:74:3)
      at processTicksAndRejections (internal/process/task_queues.js:82:21)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 on:
+  workflow_dispatch:
   push:
     branches:
       - trying
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,21 +75,6 @@ jobs:
     - name: Checkout sources
       uses: actions/checkout@v2
 
-    - name: Get npm cache directory
-      id: npm-cache
-      if: matrix.platform.cross == false
-      run: |
-        echo "::set-output name=dir::$(npm config get cache)"
-
-    - name: Cache npm (non-cross targets)
-      uses: actions/cache@v2
-      if: matrix.platform.cross == false
-      with:
-        path: ${{ steps.npm-cache.outputs.dir }}
-        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json', '**/setup.sh', '**/*.patch') }}
-        restore-keys: |
-          ${{ runner.os }}-node-
-
     - name: Install dependencies ubuntu
       if: matrix.platform.host == 'ubuntu-latest'
       run: sudo apt-get install llvm-dev libssl-dev pkg-config
@@ -138,12 +123,6 @@ jobs:
       if: matrix.platform.host == 'ubuntu-latest' && matrix.platform.cross == false
       run: cargo test --features=test_go_interop dht
 
-    - name: Setup conformance tests (non-cross targets)
-      if: matrix.platform.cross == false
-      run: ./setup.sh
-      shell: bash
-      working-directory: ./conformance
-
     - name: Build (android)
       if: contains(matrix.platform.target, 'android')
       run: cargo ndk --android-platform 29 --target ${{ matrix.platform.target }} build --workspace --exclude ipfs-http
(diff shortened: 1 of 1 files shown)


koivunej:
> ECONNRESET

these sound normal, but could be explained with something changing underneath in windows or just our tests lagging behind.

.github/workflows/ci.yml
```diff
@@ -119,85 +99,89 @@ jobs:
         rust-version: ${{ matrix.toolchain.rust }}
         targets: ${{ matrix.platform.target }}
 
-    - name: Install cargo-ndk
-      if: contains(matrix.platform.target, 'android')
-      run: cargo install --version '<2.0.0' cargo-ndk
+    - name: Rust cache
+      uses: Swatinem/rust-cache@v1
```
koivunej, line 103: Should this be given a key with the runner image name and matrix target, or can it sniff those out automatically?

.github/workflows/ci.yml
```diff
@@ -62,48 +70,20 @@ jobs:
...
       run: brew install llvm openssl
 
-    - name: Download go-ipfs on Linux
-      if: matrix.platform.host == 'ubuntu-latest'
-      run: |
-        curl -L https://github.com/ipfs/go-ipfs/releases/download/v0.7.0/go-ipfs_v0.7.0_linux-amd64.tar.gz --output go_ipfs.tar.gz
-        tar -xf go_ipfs.tar.gz
-
-    - name: Install dependencies windows (openssl)
+    - name: Install dependencies (windows)
```
koivunej, line 86: There's still the nagging about "setupOnly" which ... I don't know why it would interfere with caching. https://github.com/rs-ipfs/rust-ipfs/pull/471/checks?check_run_id=3339830233#step:33:3

.github/workflows/ci.yml
koivunej: This setup will probably be ok until we start the cross builds once more (still not convinced we should do those). Just thinking since it's part of a matrix build, perhaps it should be keyed manually. It'd seem that ubuntu and lint-rust differ only on the the prefix:

- v0-rust-lint-rust-1.54.0-x86_64-unknown-linux-gnu-a178d0322ce2-5bae624ed82b92875943
- v0-rust-ci-matrix-1.54.0-x86_64-unknown-linux-gnu-a178d0322ce2-5bae624ed82b92875943

If we'd use ubuntu to do the cross compilation into android, it would get the same cache keys.

Mirko-von-Leipzig: It uses a smart enough combination:
```
This cache is automatically keyed by:

    the github job,
    the rustc release / host / hash, and
    a hash of the Cargo.lock / Cargo.toml files (if present).
    a hash of the rust-toolchain / rust-toolchain.toml files (if present).
```

.github/workflows/ci.yml
koivunej: Ok, this is luckily nicely documented, I just had missed it: https://github.com/marketplace/actions/run-vcpkg#setup-vcpkg-only-and-use-your-own-scripts

Mirko-von-Leipzig: Just enabled this option, will test if it caches on my fork

.github/workflows/ci.yml
koivunej: If we'd enable the cross builds once more, as senseless it would seem, it would collide on ubuntu (and maybe macos), because doesn't include the target. Both ubuntu and android runs would produce competing caches as result, probably flipflopping between which gets cached.

Mirko-von-Leipzig: I was assuming that each matrix target would get its own job, but I see now that 
```
v0-rust-ci-matrix-1.54.0-x86_64-unknown-linux-gnu-a178d0322ce2-5bae624ed82b92875943
```
this implies that won't be the case. I'll add the target to the key.

Mirko-von-Leipzig: I've added `matrix.platform.target` to the cache key. Note that `key` extends the automatic key the action uses (i.e. not replace).

koivunej:
Excellent, looking great.

bors r+

.github/workflows/ci.yml
koivunej: This still requires the openssl installing step.

PR #472, opened 2021-08-17
chore: upgrade to libp2p 0.39

CHr15F0x:
This PR is a follow up of #463.

It includes the following:
- upgrading libp2p to [v.0.39.1](https://github.com/libp2p/rust-libp2p/releases/tag/v0.39.1)
- upgrading some other deps (e.g.: prost, prost-base, quick-protobuf, base64), **excluding cid and multihash**
- fixes to tests that started failing following upgrade to libp2p v.0.38 in #463

Incoming shortly:
- [x] respective changelog update
- [x] ~~I haven't run the changes against current workflow, so some additional fixes could be necessary;~~ edit: though I did check against all tests locally numerous times during dev, albeit only in a linux/x64 environment; btw looks like the workflows are fine with the changes

### DIFF
--- a/bitswap/src/protocol.rs
+++ b/bitswap/src/protocol.rs
@@ -42,7 +42,7 @@ where
     #[inline]
     fn upgrade_inbound(self, mut socket: TSocket, _info: Self::Info) -> Self::Future {
         Box::pin(async move {
-            let packet = upgrade::read_one(&mut socket, MAX_BUF_SIZE).await?;
+            let packet = upgrade::read_length_prefixed(&mut socket, MAX_BUF_SIZE).await?;
             let message = Message::from_bytes(&packet)?;
             Ok(message)
         })
@@ -71,7 +71,7 @@ where
     fn upgrade_outbound(self, mut socket: TSocket, _info: Self::Info) -> Self::Future {
         Box::pin(async move {
             let bytes = self.to_bytes();
-            upgrade::write_one(&mut socket, bytes).await
+            upgrade::write_length_prefixed(&mut socket, bytes).await
         })
     }
 }
--- a/bitswap/src/protocol.rs
+++ b/bitswap/src/protocol.rs
@@ -8,7 +8,10 @@ use crate::ledger::Message;
 use core::future::Future;
 use core::iter;
 use core::pin::Pin;
-use futures::io::{AsyncRead, AsyncWrite};
+use futures::{
+    io::{AsyncRead, AsyncWrite},
+    AsyncWriteExt,
+};
 use libp2p_core::{upgrade, InboundUpgrade, OutboundUpgrade, UpgradeInfo};
 use std::io;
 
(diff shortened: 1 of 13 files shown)


bitswap/src/protocol.rs
```diff
@@ -71,7 +71,7 @@ where
     fn upgrade_outbound(self, mut socket: TSocket, _info: Self::Info) -> Self::Future {
         Box::pin(async move {
             let bytes = self.to_bytes();
-            upgrade::write_one(&mut socket, bytes).await
+            upgrade::write_length_prefixed(&mut socket, bytes).await
```
mxinden, line 74: Note while `write_one` closes the socket, `write_length_prefixed` does not.

src/lib.rs
```diff
@@ -1368,7 +1368,8 @@ impl<TRepoTypes: RepoTypes> Future for IpfsFuture<TRepoTypes> {
         loop {
             loop {
                 let inner = {
-                    let next = self.swarm.next_event();
+                    use futures::StreamExt;
+                    let next = self.swarm.next();
```
mxinden, line 1372: Given that this is an infinite stream, you might as well use https://docs.rs/futures/0.3.16/futures/stream/trait.StreamExt.html#method.select_next_some .

mxinden (commented):
Unsolicited review. Feel free to ignore :)

bitswap/src/protocol.rs
CHr15F0x: This one I did not notice, many thanks!

src/lib.rs
CHr15F0x: Fair point, especially that indeed this is an [infinite stream](https://docs.rs/libp2p/0.39.1/libp2p/swarm/struct.ExpandedSwarm.html#impl-Stream).

src/p2p/addr.rs
```diff
@@ -215,6 +215,30 @@ pub(crate) fn could_be_bound_from_ephemeral(
...
+                } else {
+                    // All previous segments were equal
+                    return true;
+                }
+            }
+
+            true
+        }
+        _ => false,
+    }
```
Mirko-von-Leipzig, line 239: Fairly sure this is equivalent. 
```suggestion
    if addr0.is_empty() != addr1.is_empty() {
        return false;
    }

    addr0.iter().zip(addr1.iter()).all(|(a, b)| a == b)
```

CHr15F0x: Yep, this is what I've been looking for.

bitswap/src/protocol.rs
CHr15F0x: Fixed in https://github.com/rs-ipfs/rust-ipfs/pull/472/commits/87aaefd616d21316146ddd8675a925104790b7ad

src/lib.rs
CHr15F0x: https://github.com/rs-ipfs/rust-ipfs/pull/472/commits/87aaefd616d21316146ddd8675a925104790b7ad

src/p2p/addr.rs
CHr15F0x: Hopefully this function will be removed altogether in a follow-up PR which will aim at simplifying the code after peer id's been included in multiaddr in more cases.; at the moment fixed in https://github.com/rs-ipfs/rust-ipfs/pull/472/commits/87aaefd616d21316146ddd8675a925104790b7ad.

koivunej:
Cancelling the post-rebase run because bors will rerun it.

bors r+

koivunej:
Thank you @CHr15F0x for going through with this. We mostly had the "review" in chat but we mostly noted the `eq_greedy` and cleaning up of the whole swarm code in a follow-up.

PR #473, opened 2021-08-23
Swarm cleanup following libp2p upgrade to v0.39.1

CHr15F0x:
This is a follow up of https://github.com/rs-ipfs/rust-ipfs/pull/472. There are two changes introduced:
* remove eq_greedy as `Multiaddr` in `SwarmApi::pending_{addresses, connections}` contains p2p, align conversions accordingly (https://github.com/rs-ipfs/rust-ipfs/commit/136496f7de57d72782cf4c06108733321b05e9fc),
* explicitly use `MultiaddrWithPeerId` in `SwarmApi::pending_{addresses, connections}` (https://github.com/rs-ipfs/rust-ipfs/commit/3b6f69530f633a9db4545fffe4b054dc69a760fc).

The former (https://github.com/rs-ipfs/rust-ipfs/commit/136496f7de57d72782cf4c06108733321b05e9fc) works fine without the latter (https://github.com/rs-ipfs/rust-ipfs/commit/3b6f69530f633a9db4545fffe4b054dc69a760fc), however I am leaning more towards keeping both changes. Encoding the information about peer id presence/absence in Multiaddr in _type_ instead of relying on comments and `expect()`s improves readability and _developer experience_.

Please let me know what your thoughts are.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * fix: compilation error when used as a dependency [#470]
 * perf: use hash_hasher where the key is Cid [#467]
 * chore: upgrade to libp2p 0.39.1, update most of the other deps with the notable exception of cid and multihash [#472]
+* refactor(swarm): swarm cleanup following libp2p upgrade to v0.39.1 [#473]
 
 [#429]: https://github.com/rs-ipfs/rust-ipfs/pull/429
 [#428]: https://github.com/rs-ipfs/rust-ipfs/pull/428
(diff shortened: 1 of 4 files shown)


koivunej (approved):
Looking good! I think there's further simplification possible by using direct dials with a `.../p2p/PeerId` multiaddr, but this is good enough for now.

Windows test failure is again ... the transient one. Should really look into what is happening there.

bors r+

mxinden:
> I am leaning more towards keeping both changes. Encoding the information about peer id presence/absence in Multiaddr in _type_ instead of relying on comments and `expect()`s improves readability and _developer experience_.

:+1: 

I think it is even worth exploring including the change above (and thus enforcing more than just multiaddr syntax at the type level) in https://github.com/multiformats/rust-multiaddr directly.

PR #474, opened 2021-08-30
feat: (re)enable mdns

Mirko-von-Leipzig:
This PR re-enables mDNS support. This was disabled in PR #446, due to libp2p removing tokio support for it. The support has since been added back so this should work again.


Mirko-von-Leipzig:
Note that the require `libp2p` feature `mdns` brings in the `async-io` dependency

koivunej:
I haven't had much to time for rust-ipfs recently but to summarize our discussion on the current impl of `libp2p-mdns`; it still introduces unconditionally `async-io`, so the original reason for ugly commenting out has not changed. My original plan was to look into how `libp2p-tcp` was kept compatible for both runtimes. I still think we should look into restoring the functionality in upstream before re-enabling the feature over at `rust-ipfs`.

PR #479, opened 2021-12-02
Return time since connection if no round trip yet

### LINKED ISSUE #178: No latency in verbose swarm/peers before a ping response

jmmaloney4:
Fixes #178.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * fix: strict ordering for DAG-CBOR-encoded map keys [#493]
 * feat: upgrade libp2p to v0.43.0 [#499]
 * feat(http): default values for --bits and --profile [#502]
+* feat: return time since connecting began instead of null for latency [#479]
 
 [#429]: https://github.com/rs-ipfs/rust-ipfs/pull/429
 [#428]: https://github.com/rs-ipfs/rust-ipfs/pull/428
@@ -33,6 +34,7 @@
 [#493]: https://github.com/rs-ipfs/rust-ipfs/pull/493
 [#499]: https://github.com/rs-ipfs/rust-ipfs/pull/499
 [#502]: https://github.com/rs-ipfs/rust-ipfs/pull/502
+[#479]: https://github.com/rs-ipfs/rust-ipfs/pull/479
 
 # 0.2.1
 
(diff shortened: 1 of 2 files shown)


koivunej:
I gave this a quick look and it seems great, thanks for doing this! I was thinking that maybe the one of the existing hashmaps could perhaps be used for this but ... That would then complicate other things. The connection tracking parts are a bit overdue for proper cleanup and moving upwards to rust-libp2p.

There's now some npm related breakage but the build failing at that step means that the (few) rust tests passed. Oki, it's most likely because I have forgotten an old url there. I'll probably handle this on a separate PR.

Could you take care of the rustfmt issue? Just do a `cargo fmt` and commit.

jmmaloney4:
Oops

mxinden:
> The connection tracking parts are a bit overdue for proper cleanup and moving upwards to rust-libp2p.

Happy to discuss tracking the time of a connection in rust-libp2p. Feel free to open an issue on https://github.com/libp2p/rust-libp2p/.

koivunej:
Thanks for this @jmmaloney4. I have to admit I forgot about this PR while hustling the #480, apologies for that.

I've rebased this on top of the latest master. The second commit was not needed anymore so it was dropped, and I simplified with the use of [`Instant::elapsed`](https://doc.rust-lang.org/stable/std/time/struct.Instant.html#method.elapsed). I also edited the PR description to remove the checklist which does not really apply here, and it will be copied by bors to the merge commit.

It looks like conformance tests agree, so I'll push a changelog entry and merge this. Thanks again.

koivunej:
@mxinden 
> Happy to discuss tracking the time of a connection in rust-libp2p. Feel free to open an issue on https://github.com/libp2p/rust-libp2p/.

I will try to remember this and apologies on the late reply once more for you as well. I think the big cleanup is overdue with me really understanding the `Merge inject_* paired methods` PR working through these easier changes. Mostly the problems are within the rust-ipfs, trying to write code to support different interleavings of polls vs. injections while providing a futures api.

koivunej:
bors retry

koivunej:
Well, this is ackward.

bors retry

koivunej:
Well, it would appear that gha is having some issues now with the bad credentials issue. I'll try again later.

koivunej:
bors retry

Issue #483, opened 2021-12-05
Upgrade to libp2p 0.41


matthiasbeyer:
SSIA

koivunej:
> SSIA

I do not know what does this mean, care to elaborate?

matthiasbeyer:
Ah, sorry :laughing: Thats short for "subject says it all"... Sorry for the confusion!

PR #485, opened 2021-12-06
Update libp2p

matthiasbeyer:
Update libp2p to 0.40.0 and then to 0.41.0

This is my take on #483 - as I am still learning how everything works (and this seems like a non-trivial change), I'd love to get some guidance on this PR!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ either = { default-features = false, version = "1.5" }
 futures = { default-features = false, version = "0.3.9", features = ["alloc", "std"] }
 hash_hasher = "2.0.3"
 ipfs-unixfs = { version = "0.2", path = "unixfs" }
-libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-tokio", "mplex", "noise", "ping", "yamux", "dns-tokio"], version = "0.39.1" }
+libp2p = { default-features = false, features = ["floodsub", "identify", "kad", "tcp-tokio", "mplex", "noise", "ping", "yamux", "dns-tokio"], version = "0.40.0" }
 multibase = { default-features = false, version = "0.9" }
 multihash = { default-features = false, version = "0.11" }
 prost = { default-features = false, version = "0.8" }
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -377,7 +377,7 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<BitswapEvent> for Behaviour<
 
 impl<Types: IpfsTypes> NetworkBehaviourEventProcess<PingEvent> for Behaviour<Types> {
     fn inject_event(&mut self, event: PingEvent) {
-        use libp2p::ping::handler::{PingFailure, PingSuccess};
+        use libp2p::ping::{PingFailure, PingSuccess};
         match event {
             PingEvent {
                 peer,
--- a/src/p2p/pubsub.rs
+++ b/src/p2p/pubsub.rs
@@ -286,16 +286,6 @@ impl NetworkBehaviour for Pubsub {
         self.floodsub.inject_event(peer_id, connection, event)
     }
 
-    fn inject_addr_reach_failure(
-        &mut self,
-        peer_id: Option<&PeerId>,
-        addr: &Multiaddr,
-        error: &dyn std::error::Error,
-    ) {
-        self.floodsub
-            .inject_addr_reach_failure(peer_id, addr, error)
-    }
-
     fn inject_dial_failure(&mut self, peer_id: &PeerId) {
         self.floodsub.inject_dial_failure(peer_id)
     }
--- a/src/p2p/swarm.rs
+++ b/src/p2p/swarm.rs
@@ -355,37 +355,6 @@ impl NetworkBehaviour for SwarmApi {
         }
     }
 
-    fn inject_addr_reach_failure(
-        &mut self,
-        peer_id: Option<&PeerId>,
-        addr: &Multiaddr,
-        error: &dyn std::error::Error,
-    ) {
-        trace!("inject_addr_reach_failure {} {}", addr, error);
-
-        if let Some(peer_id) = peer_id {
-            match self.pending_connections.entry(*peer_id) {
-                Entry::Occupied(mut oe) => {
-                    let addresses = oe.get_mut();
-                    let addr = MultiaddrWithPeerId::try_from(addr.clone())
-                        .expect("dialed address contains peerid in libp2p 0.38");
-                    let pos = addresses.iter().position(|a| *a == addr);
-
-                    if let Some(pos) = pos {
-                        addresses.swap_remove(pos);
-                        self.connect_registry
-                            .finish_subscription(addr.into(), Err(error.to_string()));
-                    }
-
-                    if addresses.is_empty() {
-                        oe.remove();
-                    }
-                }
-                Entry::Vacant(_) => {}
-            }
-        }
-    }
-
     fn poll(
         &mut self,
         _: &mut Context,


koivunej:
Thanks for looking into this! I'll try to find time to fix the build, and I'll continue approving this. Please log if you are hitting any difficult spots, that'd be good to know for me at least.

matthiasbeyer:
So I fail to understand the issue here.

The failures are missing implementations of `NetworkBehaviour` for `Void`, but I don't understand how to resolve these, because I don't understand what these `NetworkBehaviour` things are and what changed in libp2p from 0.39.1 to 0.40.0 that makes the code no longer compile.

I would like to learn, though, so maybe you can explain or guide me on my journey?

koivunej:
I haven't yet looked what has been going on. From your mention of `Void` my knee-jerk guess is that it has been replaced by something like `std::convert::Infallible` but this is just a guess. Looking into this now. The rust-libp2p does have changelogs but they are quite difficult to follow due to their multiple files nature.

matthiasbeyer:
Yes, and the complexity of the API is so high that I completely fail to understand how everything works, especially given that I'm new to the whole topic. As said: I'd like to learn everything, though... so if you're okay with guiding me and effectively delaying this whole process (because I am learning the whole thing vs someone doing this who already knows what to do), I'd love to get this done and become more proficient during the process! :laughing:

koivunej:
I did try to look and the error message does look confusing nor did I see quickly anything from the changelogs. I need to jump on a call and such but I did notice you started the upgrade from the main/root/ipfs library crate. It might be easier (or it has been easier in the past) to first handle `bitswap/` in isolation, while preparing for bigger changes. Did not yet test if it would be applicable here.

koivunej:
Ok having now tried upgrading just `bitswap/` it is definitely more clear set of changes I didn't yet find over at rust-libp2p side. For this "workflow" I:
1. edit the `bitswap/Cargo.toml`
2. only do `cargo check --all-targets` under the `bitswap/` directory.

After fixing up `bitswap/` it's more a matter of making sure the `ipfs` crate uses the same version of `libp2p-{core,swarm}` -- there are sometimes discrepencies, like this time around [libp2p-core has changelog entry for released 0.30.1](https://github.com/libp2p/rust-libp2p/blob/master/core/CHANGELOG.md#0301-2021-11-16) but that is not found on crates.io, only 0.30.0 is. Sometimes new features could had also been split from the old ones or there could had been more drastic dependency changes.

Sadly most of the libp2p upgrades are quite a lot of work if one hasn't been following every notification from the repository, which I've long since stopped, since it was quite overwhelming.

However, we did try to add test coverage for the features we got working at one or another point, so once the rustc is satisfied you should be able to discover through the tests if any of our behavioural expectations have changed. This time around the road to that point just seems longer than usual.

Please continue logging your progress and I'll try to find some time during the holidays to also look into this and helping you.

mxinden:
:wave: rust-libp2p maintainer here. Feel free to tag me with concrete questions related to rust-libp2p and the recent breaking changes.

matthiasbeyer:
So I tried to update the `/bitswap` subtree starting fresh from `master` [here](https://github.com/matthiasbeyer/rust-ipfs/tree/update-libp2p-bitswap).

The current error I get is:

```
the trait `ProtocolsHandler` is not implemented for `BitswapEvent
```

Which I do not understand... the `BitswapEvent` is the `NetworkBehaviour::OutEvent` type. Why does this need to implement `ProtocolsHandler` at all? Isn't it just an event object?
Also, a bit counterintuitive: The type bound is only made effective via the `poll()` function, not via the traits associated type itself...

Maybe someone can educate me!? :smile:

mxinden:
The type parameters of `NetworkBehaviourAction` changed. The first parameter is now the `NetworkBehaviour::OutEvent`, the second parameter is now the `NetworkBehaviour::ProtocolsHandler`.

```rust
    events: VecDeque<NetworkBehaviourAction<Message, BitswapEvent>>,
```

should be 

```rust
    events: VecDeque<NetworkBehaviourAction<BitswapEvent, OneShotHandler<BitswapConfig, Message, MessageWrapper>>>,
```

https://github.com/matthiasbeyer/rust-ipfs/blob/ed0328bd031cde30f96d17c2ae8a4b3041d7e211/bitswap/src/behaviour.rs#L91

``` rust
        -> Poll<NetworkBehaviourAction<<<Self::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::InEvent, Self::OutEvent>>
```

should be

```rust
        -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ProtocolsHandler>>
```

https://github.com/matthiasbeyer/rust-ipfs/blob/ed0328bd031cde30f96d17c2ae8a4b3041d7e211/bitswap/src/behaviour.rs#L293

Does that help @matthiasbeyer?

matthiasbeyer:
It did!
Next step is the changed `NetworkBehaviourAction::Dial` interface, that changed as well. I assume that the `handler` that needs to be added to that object is some kind of callback mechanism. I don't know how to construct one properly though. 

Can you please tell me in a few short sentences what I am implementing here, maybe I understand what to do myself (no hurry though)!

My current code: 

* Branch: https://github.com/matthiasbeyer/rust-ipfs/tree/update-libp2p-bitswap
* Latest patch: https://github.com/matthiasbeyer/rust-ipfs/commit/0e12fed11610c9fdd38e7296143742ddbb96e6e8 (Ctrl-F TODO)

mxinden:
> I don't know how to construct one properly though.

For now you can just call `Bitswap::new_handler`.

> I assume that the handler that needs to be added to that object is some kind of callback mechanism.

A handler implements the [`ProtocolsHandler`](https://docs.rs/libp2p/0.41.0/libp2p/swarm/trait.ProtocolsHandler.html) trait. It handles all substreams of a specific protocol (e.g. bitswap) on a single connection.

> We can also switch to email or another communication channel if you want, so we do not spam this issue here! laughing

I prefer this issue as (a) others can follow along and help and (b) it might be helpful for others.

matthiasbeyer:
Ah, saw your comment too late and figured out a way to construct a new `NetworkBehaviourAction::Dial`, and pushed it to the branch linked above. Would be nice to get some comment on whether that is the right way! :smiley: 

The remaining bits is just some issue in the error type I will take care of myself.

rand0m-cloud:
Hello, I've recently started work on updating the crate to use libp2p v0.43.0 due to a build error when I try `cargo build -p ipfs-http`
```rust
error[E0282]: type annotations needed
   --> /home/xxx/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-noise-0.32.0/src/protocol/x25519.rs:221:45
    |
221 |         curve25519_sk.copy_from_slice(&hash.as_ref()[..32]);
    |                                        -----^^^^^^--
    |                                        |    |
    |                                        |    cannot infer type for type parameter `T` declared on the trait `AsRef`
    |                                        this method call resolves to `&T`
    |
    = note: type must be known at this point

For more information about this error, try `rustc --explain E0282`.
```

I got through all the type renames and then hit this error like I assume you have:
```rust
error[E0277]: the trait bound `Void: ConnectionHandler` is not satisfied
   --> src/p2p/pubsub.rs:305:15
    |
305 |     ) -> Poll<PubsubNetworkBehaviourAction> {
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ConnectionHandler` is not implemented for `Void`
    |
    = note: required because of the requirements on the impl of `IntoConnectionHandler` for `Void`
```

I scrubbed through the documentation and found that `IntoConnectionHandler` and `ConnectionHandler` added a trait bound of `Send` which `void::Void` curiously doesn't implement. I don't know how to proceed now because I guess we need to send an upstream patch to `void` to add `Sync + Send` to `void::Void`. At the same time, I would like to know why `void::Void` was in use over the `()` type.

Sorry if this is unnecessary, @mxinden any thoughts?

mxinden:
> At the same time, I would like to know why `void::Void` was in use over the `()` type.

`()` is the [unit type](https://doc.rust-lang.org/std/primitive.unit.html), thus can take on a single value, namely `()`. `Void` is the uninhabited type, i.e. it does not have any values and is thus impossible to instantiate.

PR #489, opened 2021-12-22
Gossipsub support

matthiasbeyer:
Hi!

I am new to this project and to the internals of how ipfs works - but I want to use gossipsub and this crate does not yet have support for it, from what I see. So here's my take on implementing it.

Please note that this is a very early draft - so you can stop me from doing anything idiotic before I drown myself in this! :laughing: 

Please tell me if the direction I am heading towards is right - or tell me how to do better!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -273,6 +273,11 @@ enum IpfsEvent {
     PubsubPublish(String, Vec<u8>, OneshotSender<()>),
     PubsubPeers(Option<String>, OneshotSender<Vec<PeerId>>),
     PubsubSubscribed(OneshotSender<Vec<String>>),
+    GossipsubSubscribe(String, OneshotSender<Option<SubscriptionStream>>),
+    GossipsubUnsubscribe(String, OneshotSender<bool>),
+    GossipsubPublish(String, Vec<u8>, OneshotSender<()>),
+    GossipsubPeers(Option<String>, OneshotSender<Vec<PeerId>>),
+    GossipsubSubscribed(OneshotSender<Vec<String>>),
     WantList(
         Option<PeerId>,
         OneshotSender<Vec<(Cid, ipfs_bitswap::Priority)>>,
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
No need to approve the workflow run! I just want you to either tell me that I'm writing bullshit here or that the direction is okay and what I should do to make this go into a mergeable direction! :laughing:

src/lib.rs
```diff
@@ -273,6 +273,11 @@ enum IpfsEvent {
     PubsubPublish(String, Vec<u8>, OneshotSender<()>),
     PubsubPeers(Option<String>, OneshotSender<Vec<PeerId>>),
     PubsubSubscribed(OneshotSender<Vec<String>>),
+    GossipsubSubscribe(String, OneshotSender<Option<SubscriptionStream>>),
+    GossipsubUnsubscribe(String, OneshotSender<bool>),
+    GossipsubPublish(String, Vec<u8>, OneshotSender<()>),
+    GossipsubPeers(Option<String>, OneshotSender<Vec<PeerId>>),
+    GossipsubSubscribed(OneshotSender<Vec<String>>),
```
koivunej, line 280: Last I looked, gossipsub is a superset of the floodsub so I think the old entries in IpfsEvent and generally any api just should switch to using gossipsub as the backend. At least it should spare you from writing the boilerplate methods.

src/p2p/behaviour.rs
```diff
@@ -532,6 +538,10 @@ impl<Types: IpfsTypes> Behaviour<Types> {
         &mut self.pubsub
     }
 
+    pub fn gossipsub(&mut self) -> &mut Gossipsub {
+        &mut self.gossipsub
+    }
+
```
koivunej, line 544: This would become unnecessary with the upgrade of pubsub to use gossipsub.

koivunej (changes requested):
It's a nice start, but as mentioned in the comments, just upgrading the current src/p2p/pubsub.rs from floodsub to gossipsub would be the preferred way, hopefully leaving you to do only minimal boilerplate changes.

One issue I can't remember checking with the gossipsub implementation is that are the locally sent messages possible to be echoed in subscriptions. This behaviour at least was expected by the ipfs-interface test suite (the js tests) but it is very much a "cli only" functionality which might not require changes to gossipsub, but would have to be shimmied some other way. However, that's something to worry in the later phases or next PR.

dariusc93:
Any update on this? If OP isnt interested I may try to take up on bringing gossip. Still reading over specs, libp2p code but it should be a simple implementation to replace floodsub but is there intentions to keep/use both protocols?

koivunej:
> Still reading over specs, libp2p code but it should be a simple implementation to replace floodsub but is there intentions to keep/use both protocols?

@dariusc93 No intention to keep both as gossipsub impl in libp2p handles floodsub as well. A simple implementation up to the point of the local echo or that's what I've been assuming, but I haven't checked the gossipsub code in a while.

matthiasbeyer:
I no longer have time to work on this, sorry. If someone wants to pick this up, feel free to. If not, this can of course be closed.

dariusc93:
> I no longer have time to work on this, sorry. If someone wants to pick this up, feel free to. If not, this can of course be closed.

No need to be sorry. Already picked up on this in #507 though still investigating a couple of things (some related to gossipsub and others not so much such as mdns, bootstrap, etc).

PR #490, opened 2021-12-30
bitswap: Update libp2p

matthiasbeyer:
As suggsted in #485 I re-started my efforts starting with the `/bitswap` subtree for updating libp2p.

Supersedes #485 

---

Feel free to suggest how to continue! :laughing:

### DIFF
--- a/bitswap/Cargo.toml
+++ b/bitswap/Cargo.toml
@@ -18,7 +18,7 @@ hash_hasher = "2.0.3"
 libp2p-core = { default-features = false, version = "0.30" }
 libp2p-swarm = { default-features = false, version = "0.32" }
 multihash = { default-features = false, version = "0.11" }
-prost = { default-features = false, version = "0.8" }
+prost = { default-features = false, version = "0.9" }
 thiserror = { default-features = false, version = "1.0" }
 tokio = { default-features = false, version = "1", features = ["rt"] }
 tracing = { default-features = false, version = "0.1" }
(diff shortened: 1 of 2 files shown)


koivunej:
Great you've been making progress, thanks also for Max (cannot tag from mobile ui) for helping out. I will be out of office until next week (xmas holidays) when I hope to fix the build and then merge this. Again, thanks for doing this!

Note that I only suggested that you'd start the upgrade from first getting the `bitswap/` work, possibly leaving it implicit that you should continue to upgrade `ipfs` or the root crate after that. The libp2p upgrade should be on a single PR, even though it affects two parts or more parts (`http/`). Also, feel free to rebase your work on the PR as much as you want, just avoiding merges in the PR branch. However if the git-fu becomes an issue, I can take care of that when merging as well.

matthiasbeyer:
> I will be out of office until next week 

Sure, no hurry!

> Note that I only suggested that you'd start the upgrade from first getting the bitswap/ work, possibly leaving it implicit that you should continue to upgrade ipfs or the root crate after that.

Yes, sure! Upgrading bitswap was ... doable, but I fail to see why the root crate does not compile, so I figured I rather push this into a PR and we go from there instead me just becoming desperate :laughing: 

> However if the git-fu becomes an issue, I can take care of that when merging as well.

No worries, I am a git ninja (>10 years experience, with different workflows and even email workflows etc etc)!

mxinden (approved):
Changes itself look good to me.

koivunej (approved):
This is looking good; compilation fails on your branch though, so the rest of the upgrade needs to be done on the ipfs side. If libp2p has already advanced to a next major version, then I'd suggest doing the ongoing upgrade first before moving on to the next.

koivunej:
Similar to #485, closing this as libp2p has been upgraded to 0.43.

Issue #492, opened 2022-01-26
DAG-CBOR-encoded Ipld::Map does not strictly order keys


cdata:
**Describe the bug**
TL;DR `Ipld::Map` does not strictly order keys, leading to inconsistency in content hashing between this implementation and `go-ipfs` (and probably others).

Strictness errata for DAG-CBOR, including explicit key order, is detailed here: https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#:~:text=The%20keys%20in,order%20sorts%20earlier.

**To Reproduce**
```rust
use ipfs::{ipld::dag_cbor::DagCborCodec, Ipld};
use std::collections::BTreeMap;

fn main() {
    let mut map = BTreeMap::<String, Ipld>::new();
    map.insert("foo".into(), Ipld::Null);
    map.insert("calcium".into(), Ipld::Null);
    map.insert("zesty".into(), Ipld::Null);

    let ipld_map = Ipld::Map(map);
    let bytes = DagCborCodec::encode(&ipld_map);

    println!("Bytes: {:?}", bytes);
}
```

Output with current (incorrect) implementation:

```
Ok([163, 103, 99, 97, 108, 99, 105, 117, 109, 246, 99, 102, 111, 111, 246, 101, 122, 101, 115, 116, 121, 246])
```

**Expected behavior**

Expected output (created with patched implementation):

```
Ok([163, 99, 102, 111, 111, 246, 101, 122, 101, 115, 116, 121, 246, 103, 99, 97, 108, 99, 105, 117, 109, 246])
```

**Environment (please complete the following information)**
- Operating system, kernel version where applicable: Pop!_OS 21.10 / 5.15.11-76051511-generic
- Rust version: 

stable-x86_64-unknown-linux-gnu (default)
rustc 1.57.0 (f1edd0429 2021-11-29)

vmx:
FYI: there are also test vectors for DAG-CBOR at https://github.com/ipld/codec-fixtures. I'm currently adding a test runner to run them against another IPLD implementation: https://github.com/ipld/codec-fixtures/tree/add-rust

koivunej:
Keeping this issue open for a less wasteful implementation which would use a btreemap with correctly sorting keys newtype.

PR #493, opened 2022-01-26
Strict ordering for DAG-CBOR-encoded map keys

cdata:
This change proposes a fix for #492 . `Ipld::Map` keys are sorted according to [the strictness guidelines in the DAG-CBOR spec](https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#:~:text=The%20keys%20in,order%20sorts%20earlier.) when encoding them as DAG-CBOR.

### DIFF
--- a/src/ipld/dag_cbor.rs
+++ b/src/ipld/dag_cbor.rs
@@ -4,7 +4,6 @@ use crate::ipld::{BlockError, Ipld, IpldError};
 use byteorder::{BigEndian, ByteOrder};
 use cid::Cid;
 use std::{
-    cmp::Ordering,
     collections::BTreeMap,
     convert::TryFrom,
     io::{Read, Write},
--- a/src/ipld/dag_cbor.rs
+++ b/src/ipld/dag_cbor.rs
@@ -325,15 +325,7 @@ impl<T: WriteCbor + 'static> WriteCbor for BTreeMap<String, T> {
         let mut keys: Vec<&String> = self.keys().collect();
 
         // See: https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#strictness
-        keys.sort_by(|l, r| {
-            if r.len() > l.len() {
-                Ordering::Less
-            } else if r.len() < l.len() {
-                Ordering::Greater
-            } else {
-                l.partial_cmp(r).or(Some(Ordering::Equal)).unwrap()
-            }
-        });
+        keys.sort_by(|l, r| l.len().cmp(&r.len()).then_with(|| l.cmp(r)));
 
         for k in keys {
             let v = self.get(k).unwrap();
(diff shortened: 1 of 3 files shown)


koivunej:
Thanks for reporting and offering a fix for this. The build is currently broken, will try to get to fixing that in near future.

I'm thinking the eventual solution is to have a newtype and comparator which has the correct sort.. Do I understand this correctly, that `String` or `Vec` or [`[u8]` would sort](https://github.com/rust-lang/rust/blob/bb549e5afe8f23258606649cc3a2d8b8a51ef63a/library/core/src/slice/cmp.rs#L103-L121) correctly except that it doesn't consider the length as the first sort?

koivunej (approved):

koivunej:
Maybe it's because of RFC 8949 includes the sorting on the cbor serialized keys which start with the length information. Now it makes sense.

@cdata could you rebase this on top of master? The build was broken but now it should work.

cdata:
> Thanks for reporting and offering a fix for this. The build is currently broken, will try to get to fixing that in near future.
> 
> I'm thinking the eventual solution is to have a newtype ~and comparator~ with expected PartialOrd implementation.. Do I understand this correctly, that `String` or `Vec` or [`[T]` would sort](https://github.com/rust-lang/rust/blob/bb549e5afe8f23258606649cc3a2d8b8a51ef63a/library/core/src/slice/cmp.rs#L103-L121) correctly except that it doesn't consider the length as the first sort?

Yes, I believe that analysis of the problem is correct.

The change has been rebased onto the latest `master` version.

src/ipld/dag_cbor.rs
```diff
@@ -321,10 +322,25 @@ impl<T: WriteCbor + 'static> WriteCbor for BTreeMap<String, T> {
...
+
+        // See: https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#strictness
+        keys.sort_by(|l, r| {
+            if r.len() > l.len() {
+                Ordering::Less
+            } else if r.len() < l.len() {
+                Ordering::Greater
+            } else {
+                l.partial_cmp(r).or(Some(Ordering::Equal)).unwrap()
+            }
```
koivunej, line 335:
```suggestion
            r.len().cmp(l.len()).then_with(|| l.cmp(r))
```

Maybe clippy might be satisfied with this?

src/ipld/dag_cbor.rs
```diff
@@ -321,10 +322,25 @@ impl<T: WriteCbor + 'static> WriteCbor for BTreeMap<String, T> {
...
+        });
```
koivunej, line 336:
```suggestion
        keys.sort_by(|l, r| l.len().cmp(&r.len()).then_with(|| l.cmp(r)));
```

This works.

src/ipld/dag_cbor.rs
```diff
@@ -4,6 +4,7 @@ use crate::ipld::{BlockError, Ipld, IpldError};
 use byteorder::{BigEndian, ByteOrder};
 use cid::Cid;
 use std::{
+    cmp::Ordering,
```
koivunej, line 7:
```suggestion
```

If/when you simplify the if-elseif-else away with my suggestion, this becomes unnecessary and will most likely fail compilation.

koivunej:
Thank you for the fix!

bors r+

PR #498, opened 2022-02-28
Expose Floodsub target_peer list.

cobward:
The IPFS facade does not have access to the functions which add and remove nodes to the floodsub target_peer list. As such, messages are never propagated to connected peers.

This PR adds two methods to the IPFS facade so that these functions can be used, and so that pubsub can be used in a network of more than one IPFS peer.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -268,6 +268,8 @@ enum IpfsEvent {
     Disconnect(MultiaddrWithPeerId, Channel<()>),
     /// Request background task to return the listened and external addresses
     GetAddresses(OneshotSender<Vec<Multiaddr>>),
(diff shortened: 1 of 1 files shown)


koivunej:
Hi @cobward! Thanks for the PR. I've missed this, and now that libp2p was upgraded to 0.43 in #499 I think this is handled "automatically". Could you check and see if the automation matches what you are after?

My initial understanding is that this is something which should be handled behind the scenes and not be exposed, and I don't think this is exposed in go-ipfs ([http api](https://docs.ipfs.io/reference/http/api/#api-v0-pubsub-ls) at least). I have to admit I'm not up to speed on topic discovery for gossipsub/floodsub, and doubt the partial view adding done with the upgrade is enough for that.

PR #499, opened 2022-03-06
Update libp2p to v0.43.0

rand0m-cloud:
This PR updates the dependency on libp2p. This is needed because currently building `ipfs-http` fails with an ambiguous type error inside of `libp2p`.

### DIFF
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -14,7 +14,7 @@ use futures::channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
 use hash_hasher::HashedMap;
 use libp2p_core::{connection::ConnectionId, ConnectedPoint, Multiaddr, PeerId};
 use libp2p_swarm::dial_opts::{DialOpts, PeerCondition};
-use libp2p_swarm::handler::{IntoConnectionHandler, OneShotHandler};
+use libp2p_swarm::handler::OneShotHandler;
 use libp2p_swarm::{NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters};
 use std::task::{Context, Poll};
 use std::{
@@ -315,12 +315,7 @@ impl NetworkBehaviour for Bitswap {
         &mut self,
         ctx: &mut Context,
         _: &mut impl PollParameters,
-    ) -> Poll<
-        NetworkBehaviourAction<
-            Self::OutEvent,
-            <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
-        >,
-    > {
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
         use futures::stream::StreamExt;
 
         while let Poll::Ready(Some((peer_id, block))) = self.ready_blocks.poll_next_unpin(ctx) {
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -12,7 +12,7 @@ use cid::Cid;
 use fnv::FnvHashSet;
 use futures::channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
 use hash_hasher::HashedMap;
-use libp2p_core::{connection::ConnectionId, Multiaddr, PeerId};
+use libp2p_core::{connection::ConnectionId, ConnectedPoint, Multiaddr, PeerId};
 use libp2p_swarm::dial_opts::{DialOpts, PeerCondition};
 use libp2p_swarm::handler::{IntoConnectionHandler, OneShotHandler};
 use libp2p_swarm::{NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters};
@@ -228,6 +228,35 @@ impl NetworkBehaviour for Bitswap {
     fn addresses_of_peer(&mut self, _peer_id: &PeerId) -> Vec<Multiaddr> {
         debug!("bitswap: addresses_of_peer");
         Vec::new()
+    }
+
+    fn inject_connection_established(
+        &mut self,
+        peer_id: &PeerId,
+        _connection_id: &ConnectionId,
+        _endpoint: &ConnectedPoint,
+        _failed_addresses: Option<&Vec<Multiaddr>>,
+        _other_established: usize,
+    ) {
+        debug!("bitswap: inject_connected {}", peer_id);
+        let ledger = Ledger::new();
+        self.stats.entry(*peer_id).or_default();
+        self.connected_peers.insert(*peer_id, ledger);
+        self.send_want_list(*peer_id);
+    }
+
+    fn inject_connection_closed(
+        &mut self,
+        peer_id: &PeerId,
+        _connection_id: &ConnectionId,
+        _endpoint: &ConnectedPoint,
+        _handler: Self::ConnectionHandler,
+        _remaining_established: usize,
+    ) {
+        debug!("bitswap: inject_disconnected {:?}", peer_id);
+        self.connected_peers.remove(peer_id);
+        // the related stats are not dropped, so that they
+        // persist for peers regardless of disconnects
     }
 
     fn inject_event(&mut self, source: PeerId, _connection: ConnectionId, message: MessageWrapper) {
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -13,10 +13,9 @@ use fnv::FnvHashSet;
 use futures::channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
 use hash_hasher::HashedMap;
 use libp2p_core::{connection::ConnectionId, Multiaddr, PeerId};
-use libp2p_swarm::protocols_handler::{IntoProtocolsHandler, OneShotHandler, ProtocolsHandler};
-use libp2p_swarm::{
-    DialPeerCondition, NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters,
-};
+use libp2p_swarm::dial_opts::{DialOpts, PeerCondition};
+use libp2p_swarm::handler::{IntoConnectionHandler, OneShotHandler};
+use libp2p_swarm::{NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters};
 use std::task::{Context, Poll};
 use std::{
     collections::{HashMap, VecDeque},
@@ -88,7 +87,13 @@ impl Stats {
 /// Network behaviour that handles sending and receiving IPFS blocks.
 pub struct Bitswap {
     /// Queue of events to report to the user.
-    events: VecDeque<NetworkBehaviourAction<Message, BitswapEvent>>,
+    events: VecDeque<
+        NetworkBehaviourAction<
+            BitswapEvent,
+            <Bitswap as NetworkBehaviour>::ConnectionHandler,
+            Message,
+        >,
+    >,
     /// List of prospect peers to connect to.
     target_peers: FnvHashSet<PeerId>,
     /// Ledger
@@ -150,9 +155,11 @@ impl Bitswap {
     /// Called from Kademlia behaviour.
     pub fn connect(&mut self, peer_id: PeerId) {
         if self.target_peers.insert(peer_id) {
-            self.events.push_back(NetworkBehaviourAction::DialPeer {
-                peer_id,
-                condition: DialPeerCondition::Disconnected,
+            self.events.push_back(NetworkBehaviourAction::Dial {
+                opts: DialOpts::peer_id(peer_id)
+                    .condition(PeerCondition::Disconnected)
+                    .build(),
+                handler: todo!(),
             });
         }
     }
@@ -209,10 +216,10 @@ impl Bitswap {
 }
 
 impl NetworkBehaviour for Bitswap {
-    type ProtocolsHandler = OneShotHandler<BitswapConfig, Message, MessageWrapper>;
+    type ConnectionHandler = OneShotHandler<BitswapConfig, Message, MessageWrapper>;
     type OutEvent = BitswapEvent;
 
-    fn new_handler(&mut self) -> Self::ProtocolsHandler {
+    fn new_handler(&mut self) -> Self::ConnectionHandler {
         debug!("bitswap: new_handler");
         Default::default()
     }
@@ -220,21 +227,6 @@ impl NetworkBehaviour for Bitswap {
     fn addresses_of_peer(&mut self, _peer_id: &PeerId) -> Vec<Multiaddr> {
         debug!("bitswap: addresses_of_peer");
         Vec::new()
-    }
-
-    fn inject_connected(&mut self, peer_id: &PeerId) {
-        debug!("bitswap: inject_connected {}", peer_id);
-        let ledger = Ledger::new();
-        self.stats.entry(*peer_id).or_default();
-        self.connected_peers.insert(*peer_id, ledger);
-        self.send_want_list(*peer_id);
-    }
-
-    fn inject_disconnected(&mut self, peer_id: &PeerId) {
-        debug!("bitswap: inject_disconnected {:?}", peer_id);
-        self.connected_peers.remove(peer_id);
-        // the related stats are not dropped, so that they
-        // persist for peers regardless of disconnects
     }
 
     fn inject_event(&mut self, source: PeerId, _connection: ConnectionId, message: MessageWrapper) {
@@ -289,9 +281,16 @@ impl NetworkBehaviour for Bitswap {
     }
 
     #[allow(clippy::type_complexity)]
-    fn poll(&mut self, ctx: &mut Context, _: &mut impl PollParameters)
-        -> Poll<NetworkBehaviourAction<<<Self::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::InEvent, Self::OutEvent>>
-    {
+    fn poll(
+        &mut self,
+        ctx: &mut Context,
+        _: &mut impl PollParameters,
+    ) -> Poll<
+        NetworkBehaviourAction<
+            Self::OutEvent,
+            <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
+        >,
+    > {
         use futures::stream::StreamExt;
 
         while let Poll::Ready(Some((peer_id, block))) = self.ready_blocks.poll_next_unpin(ctx) {
--- a/bitswap/src/behaviour.rs
+++ b/bitswap/src/behaviour.rs
@@ -155,11 +155,12 @@ impl Bitswap {
     /// Called from Kademlia behaviour.
     pub fn connect(&mut self, peer_id: PeerId) {
         if self.target_peers.insert(peer_id) {
+            let handler = self.new_handler();
             self.events.push_back(NetworkBehaviourAction::Dial {
                 opts: DialOpts::peer_id(peer_id)
                     .condition(PeerCondition::Disconnected)
                     .build(),
-                handler: todo!(),
+                handler,
             });
         }
     }
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -84,7 +84,7 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<KademliaEvent> for Behaviour
         };
 
         match event {
-            InboundRequestServed { request } => {
+            InboundRequest { request } => {
                 trace!("kad: inbound {:?} request handled", request);
             }
             OutboundQueryCompleted { result, id, .. } => {
@@ -377,7 +377,7 @@ impl<Types: IpfsTypes> NetworkBehaviourEventProcess<BitswapEvent> for Behaviour<
 
 impl<Types: IpfsTypes> NetworkBehaviourEventProcess<PingEvent> for Behaviour<Types> {
     fn inject_event(&mut self, event: PingEvent) {
-        use libp2p::ping::handler::{PingFailure, PingSuccess};
+        use libp2p::ping::{PingFailure, PingSuccess};
         match event {
             PingEvent {
                 peer,
--- a/src/p2p/behaviour.rs
+++ b/src/p2p/behaviour.rs
@@ -23,7 +23,7 @@ use tokio::task;
 
 /// Behaviour type.
 #[derive(libp2p::NetworkBehaviour)]
-#[behaviour(out_event = "BehaviourEvent")]
+#[behaviour(out_event = "BehaviourEvent", event_process = true)]
 pub struct Behaviour<Types: IpfsTypes> {
     #[behaviour(ignore)]
     repo: Arc<Repo<Types>>,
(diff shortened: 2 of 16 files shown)


rand0m-cloud:
```rust
error[E0053]: method `poll` has an incompatible type for trait
   --> src/p2p/pubsub.rs:328:10
    |
328 |     ) -> Poll<PubsubNetworkBehaviourAction> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |          |
    |          expected enum `Void`, found enum `floodsub::layer::InnerMessage`
    |          help: change the output type to match the trait: `Poll<libp2p::libp2p_swarm::NetworkBehaviourAction<Void, OneShotHandler<FloodsubProtocol, FloodsubRpc, floodsub::layer::InnerMessage>, FloodsubRpc>>`
    |
    = note: expected fn pointer `fn(&mut Pubsub, &mut std::task::Context<'_>, &mut impl PollParameters) -> Poll<libp2p::libp2p_swarm::NetworkBehaviourAction<Void, _, _>>`
               found fn pointer `fn(&mut Pubsub, &mut std::task::Context<'_>, &mut impl PollParameters) -> Poll<libp2p::libp2p_swarm::NetworkBehaviourAction<floodsub::layer::InnerMessage, _, _>>`
```
I'm not sure how to approach this error.

rand0m-cloud:
@mxinden I believe these are issues stemming from libp2p.
```rust
error[E0599]: no method named `into_inner` found for struct `KademliaHandler` in the current scope
  --> src/p2p/behaviour.rs:25:10
   |
25 | #[derive(libp2p::NetworkBehaviour)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `KademliaHandler<QueryId>`
   |
   = note: this error originates in the derive macro `libp2p::NetworkBehaviour` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `into_inner` found for struct `KademliaHandlerProto` in the current scope
  --> src/p2p/behaviour.rs:25:10
   |
25 | #[derive(libp2p::NetworkBehaviour)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `KademliaHandlerProto<QueryId>`
   |
   = note: this error originates in the derive macro `libp2p::NetworkBehaviour` (in Nightly builds, run with -Z macro-backtrace for more info)
```

rand0m-cloud:
https://github.com/libp2p/rust-libp2p/blob/a168410dbed0d0941f2e5a14543206044ccb2260/swarm/Cargo.toml#L6

Isn't libp2p-swarm supposed to be version 0.35.0? 
```rust
error: failed to select a version for the requirement `libp2p-swarm = "^0.35"`
candidate versions found which didn't match: 0.34.0, 0.33.0, 0.32.0, ...
location searched: crates.io index
```

mxinden:
> https://github.com/libp2p/rust-libp2p/blob/a168410dbed0d0941f2e5a14543206044ccb2260/swarm/Cargo.toml#L6
> 
> Isn't libp2p-swarm supposed to be version 0.35.0?
> 
> ```rust
> error: failed to select a version for the requirement `libp2p-swarm = "^0.35"`
> candidate versions found which didn't match: 0.34.0, 0.33.0, 0.32.0, ...
> location searched: crates.io index
> ```

`v0.35.0` is not yet released, i.e. only in `master` thus far.

bitswap/src/behaviour.rs
```diff
@@ -222,21 +230,6 @@ impl NetworkBehaviour for Bitswap {
...
-        self.send_want_list(*peer_id);
-    }
-
-    fn inject_disconnected(&mut self, peer_id: &PeerId) {
-        debug!("bitswap: inject_disconnected {:?}", peer_id);
-        self.connected_peers.remove(peer_id);
-        // the related stats are not dropped, so that they
-        // persist for peers regardless of disconnects
-    }
-
```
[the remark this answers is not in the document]
rand0m-cloud: Ah yes, those are provided methods. Resolved! Thanks!

koivunej:
Thanks for doing this @rand0m-cloud. I was on a vacation but back now. Let me know when this is ready, or if you need any help with it.

rand0m-cloud:
I've had a chance to play around more with the code and I can't figure out to how to solve the error. I've compared it to the `file-sharing` example of libp2p, and I can't make sense of what is wrong. The error is from the derive macro `libp2p::NetworkBehaviour` expecting some handler to implement `into_inner`, but it's my understanding that `into_inner` only exists for the `ConnectionHandlerSelect` that the macro creates.

I am lost but it feels like a bug in the derive macro.

mxinden:
@rand0m-cloud mind posting the compiler error here?

rand0m-cloud:
Sure, the errors in the PR description are the current ones but here it is again:
```rust
error[E0599]: no method named `into_inner` found for struct `KademliaHandler` in the current scope
  --> src/p2p/behaviour.rs:25:10
   |
25 | #[derive(libp2p::NetworkBehaviour)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `KademliaHandler<QueryId>`
   |
   = note: this error originates in the derive macro `libp2p::NetworkBehaviour` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `into_inner` found for struct `KademliaHandlerProto` in the current scope
  --> src/p2p/behaviour.rs:25:10
   |
25 | #[derive(libp2p::NetworkBehaviour)]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `KademliaHandlerProto<QueryId>`
   |
   = note: this error originates in the derive macro `libp2p::NetworkBehaviour` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
```

bitswap/src/behaviour.rs
```diff
@@ -289,9 +311,16 @@ impl NetworkBehaviour for Bitswap {
...
-    {
+    fn poll(
+        &mut self,
+        ctx: &mut Context,
+        _: &mut impl PollParameters,
+    ) -> Poll<
+        NetworkBehaviourAction<
+            Self::OutEvent,
+            <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
+        >,
```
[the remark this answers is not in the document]
rand0m-cloud: No, it didn't resolve the issue. I added that commit (and hope the types were equivalent because it would confuse me even more). Just the `ipfs` crate has a build issue.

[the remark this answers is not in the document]
rand0m-cloud: I would run a cargo expand on the crate but I know far too little about this crate and libp2p to be helpful.

src/p2p/behaviour.rs
```diff
@@ -15,13 +15,15 @@ use libp2p::kad::{Kademlia, KademliaConfig, KademliaEvent, Quorum};
...
 // use libp2p::swarm::toggle::Toggle;
+use libp2p::floodsub::FloodsubEvent;
 use libp2p::swarm::{NetworkBehaviour, NetworkBehaviourEventProcess};
 use multibase::Base;
 use std::{convert::TryInto, sync::Arc};
 use tokio::task;
 
 /// Behaviour type.
 #[derive(libp2p::NetworkBehaviour)]
+#[behaviour(out_event = "BehaviourEvent", event_process = true)]
```
mxinden, line 26:
```suggestion
#[behaviour(event_process = true)]
```

Why is the `OutEvent` needed here? Given that you set `event_process = true` there is no need for it, right?

src/p2p/behaviour.rs
```diff
@@ -15,13 +15,15 @@ use libp2p::kad::{Kademlia, KademliaConfig, KademliaEvent, Quorum};
...
 pub struct Behaviour<Types: IpfsTypes> {
```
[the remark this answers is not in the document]
rand0m-cloud: Yep! That was it!

rand0m-cloud:
Thanks to @mxinden's workaround, I was able to continue working on the PR. Now the PR builds, lints, but fails testing:
```rust

failures:

---- p2p::swarm::tests::racy_connecting_attempts stdout ----
thread 'p2p::swarm::tests::racy_connecting_attempts' panicked at 'assertion failed: `(left == right)`
  left: `[Err(Some("addresses exhausted")), Err(Some("addresses exhausted"))]`,
 right: `[Ok(()), Err(Some("finished connecting to another address"))]`', src/p2p/swarm.rs:517:21

---- p2p::swarm::tests::wrong_peerid stdout ----
thread 'p2p::swarm::tests::wrong_peerid' panicked at 'assertion failed: `(left == right)`
  left: `Some("addresses exhausted")`,
 right: `Some("Pending connection: Invalid peer ID.")`', src/p2p/swarm.rs:466:21


failures:
    p2p::swarm::tests::racy_connecting_attempts
    p2p::swarm::tests::wrong_peerid

test result: FAILED. 85 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.66s
```

src/p2p/behaviour.rs
rand0m-cloud: Resolved!

src/p2p/behaviour.rs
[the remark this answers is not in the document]
mxinden: Thanks for the feedback. Tracked here: https://github.com/libp2p/rust-libp2p/issues/2569

mxinden:
> Thanks to @mxinden's workaround, I was able to continue working on the PR. Now the PR builds, lints, but fails testing:

Sorry, but I don't think I am of much help on the rust-ipfs specific tests.

koivunej:
@rand0m-cloud Yeah ... these connection tests ... They are most likely extra, and become more and more useless now that libp2p must've evolved. Back in the day it was a bit more tricky trying to get the similar behaviour as to go-ipfs. I'll try to check the branch locally tomorrow! Thanks for your work so far.

src/p2p/swarm.rs
```diff
@@ -335,53 +347,35 @@ impl NetworkBehaviour for SwarmApi {
...
+            // this should not be executed once, but probably will be in case unsupported addresses or something
+            // surprising happens.
+            for failed in self
+                .pending_connections
+                .remove(&peer_id)
+                .unwrap_or_default()
+            {
+                self.connect_registry
+                    .finish_subscription(failed.into(), Err("addresses exhausted".into()));
             }
```
koivunej, line 379: Doing this unconditionally creates the failure for `p2p::swarm::tests::racy_connecting_attempts`.

I wonder ... why were changes around this necessary? Seems this logic would had been easiest to do handle with the old structure using the entry api?

koivunej:
So in total I see 3 test failures:
- `swarm_api` (really bad name)
- `racy_connection_attempts`
- `wrong_peerid`

For the last two, I think a solution should come out of the discussion thread I already started.

For the first one, I think the near-correct answer is to add `biased;` to the `tokio::select!` so the events are observed in the assumed order always. Though, this is prone to break because we are racing with the idle disconnection timer. However, I remember looking at the idle disconnection should be done right away so this might be ok. I did not yet push this fix, let me know if it should be pushed. If this doesn't fail for you, please let me know!

koivunej (approved):
Apart from my comment on the inject_dial_failure I think this is looking good!

rand0m-cloud:
@koivunej Okay, I think I've re-added the code fragment I removed and would make `inject_dial_failure` work correctly.

The problem is the original code was given the Multiaddr it failed to dial from `inject_addr_reach_failure` and the new `inject_dial_failure` doesn't provide that. I can't find a clear way to go from the `PeerId` to `MultiaddrWithPeerId`.

rand0m-cloud:
Do we need a new `ConnectionHandler` that can store the Multiaddr?

mxinden:
> The problem is the original code was given the Multiaddr it failed to dial from `inject_addr_reach_failure` and the new `inject_dial_failure` doesn't provide that. I can't find a clear way to go from the `PeerId` to `MultiaddrWithPeerId`.

It does, though instead of calling `inject_addr_reach_failure` for each address of a single dial, it provides all the failed addresses at once via `DialError::Transport` on failure of all addresses in `inject_dial_failure` and a `Vec` of addresses that failed, in case a single one succeeded via `inject_connection_established` `failed_addresses`.

https://docs.rs/libp2p/latest/libp2p/swarm/enum.DialError.html

https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html#method.inject_connection_established

src/p2p/swarm.rs
```diff
@@ -335,36 +347,29 @@ impl NetworkBehaviour for SwarmApi {
...
+        error: &DialError,
     ) {
-        trace!("inject_addr_reach_failure {} {}", addr, error);
+        trace!("inject_dial_failure: {:?} ({})", peer_id, error);
+        if let Some(peer_id) = peer_id {
+            if self.pending_addresses.contains_key(&peer_id) {
+                // it is possible that these addresses have not been tried yet; they will be asked
+                // for soon.
+                let handler = self.new_handler();
+                self.events.push_back(swarm::NetworkBehaviourAction::Dial {
```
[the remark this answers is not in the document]
koivunej: All right ... Looking at the networkbehaviour, I think this is how it used to work:

1. dial event is given to swarm
2. swarm collects all of the known addresses with `NetworkBehaviour::addresses_of_peer`, dials them somehow
3. for each of the dials, we used to get a signal that this multiaddr failed and we would signal that future as failure
4. there used to be another trait method for having exhausted the addresses, when we'd know to launch a new dial if we'd still have addresses

While writing this @mxinden [replied](https://github.com/rs-ipfs/rust-ipfs/pull/499#issuecomment-1073724116). Oki yeah it would appear the failures have now moved within the dialerror, which I did not expect, AND there is only one "notification" for all of the attempts.

However I think the idea with the original impl was that since there would be one gathering of addresses for the peer (`NetworkBehaviour::addresses_of_peer`) per dial events (caused either by "swarm_api" or by any other place) it would be possible to add new addresses during a dial attempt, and those would be noticed at (4) in the above ordered list, and thus get dialed afterwards.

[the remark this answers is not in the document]
koivunej: I think for the next steps would be to gather the failed addresses from the error (if found), then continue dialing to the remaining addresses, if any.

rand0m-cloud:
Great! Solved the test failures and changed the dial failure logic to use DialError. Now this pubsub test is failing:

```rust
failures:

---- publish_between_two_nodes stdout ----
thread 'publish_between_two_nodes' panicked at 'called `Result::unwrap()` on an `Err` value: Elapsed(())', tests/pubsub.rs:120:10


failures:
    publish_between_two_nodes

```

src/p2p/swarm.rs
```diff
@@ -310,71 +324,51 @@ impl NetworkBehaviour for SwarmApi {
...
+                                    MultiaddrWithPeerId::try_from(addr.clone()).unwrap()
+                                })
+                                .collect::<Vec<_>>();
+
+                            addresses.retain(|peer_id| peer_ids.iter().any(|id| peer_id == id));
+                        }
+                        DialError::WrongPeerId {
+                            obtained: _,
+                            endpoint: _,
+                        } => {
```
rand0m-cloud, line 361: Do we have a preference on `Foo { a: _ }` verus `Foo { .. }`

src/p2p/swarm.rs
```diff
@@ -310,71 +324,51 @@ impl NetworkBehaviour for SwarmApi {
...
+                            for addr in addresses.iter() {
+                                self.connect_registry.finish_subscription(
+                                    addr.clone().into(),
+                                    Err(error.to_string()),
+                                );
+                            }
+
+                            addresses.clear();
+                        }
+                        err => trace!("unhandled DialError {}", err),
```
rand0m-cloud, line 371: Should we trace every DialError? Was thinking about the WrongPeerId failure above and I'm not sure if it would otherwise be reported to the user.

src/p2p/swarm.rs
koivunej: I guess I'll usually just write `..` as it's less typing :)

src/p2p/swarm.rs
koivunej: I would probably go with

```suggestion
                        error => {
                          warn!(?error, "unexpected DialError; some futures might never complete");               
                        },
```

Unless you can see a way to finish all of the subscriptions under this peerid? I cannot at least, we had not considered this situation originally as mentioned the events used to all arrive.

src/p2p/swarm.rs
```diff
@@ -310,71 +324,51 @@ impl NetworkBehaviour for SwarmApi {
...
+                        DialError::Transport(multiaddrs) => {
+                            for (addr, error) in multiaddrs {
+                                let addr = MultiaddrWithPeerId::try_from(addr.clone())
+                                    .expect("to recieve an MultiAddrWithPeerId from DialError");
+                                self.connect_registry
+                                    .finish_subscription(addr.into(), Err(error.to_string()));
+                            }
+
+                            let peer_ids = multiaddrs
+                                .into_iter()
```
koivunej, line 350:
```suggestion
                                .iter()
```

The error is behind a borrow so cannot take ownership here. I was going to ask about the `addr.clone()` on line 352 but this explains it.

However ... This vec building seems extra because we are already handling a failure to dial a specific peer, and we only allow connect's to `MultiaddrWithPeerId` on line 97 with `pub fn connect(&mut self, addr: MultiaddrWithPeerId) -> ...`.

So perhaps it would be better to remove from `addresses` with in the `finish_subscription` loop above. It will need to be position based swap remove.

rand0m-cloud:
I'm glad I caught my mistake before I logged off, but now its time to take a much-needed break.

(the build times are painful, think I remember seeing incremental builds being disabled on stable)

koivunej:
Yeah the build times are further exacerbated by having many of the `tests/` which require linking and everything is going very slowly. I haven't yet tried [mold](https://github.com/rui314/mold) with the stable with forced incremental builds (can't remember that env variable either).

src/p2p/swarm.rs
mxinden: My two cents:

- `Foo { a: _ }` when you expect there to be new fields to be added in the future which you would like to handle and thus want compilation to fail.

- `Foo { .. }` if you don't care about the fields now and the future.

rand0m-cloud:
So, I'm trying to work through the pubsub failures and the following methods are missing from the NetworkBehaviour impl on Pubsub:
```
inject_address_change
inject_listen_failure
inject_expired_external_addr
inject_listener_closed
inject_new_listener
```

I was hoping to migrate it to use the NetworkBehaviour derive macro but the custom poll function makes that seem non trivial. I will work on restoring these methods and see if that resolves the test.

rand0m-cloud:
Changing the expected messages received to one reveals that the test is hanging waiting for the second message to arrive.
```rust
---- publish_between_two_nodes stdout ----
thread 'publish_between_two_nodes' panicked at 'assertion failed: `(left == right)`
  left: `{(["shared"], PeerId("12D3KooWJfVQNYbTgmAjfzsMiWANjLdh225jMfG3VH2hCA5xDBDc"), [102, 111, 111, 98, 97, 114]), (["shared"], PeerId("12D3KooWRW9MXWPG9VaFFbyPWxCfZw4geE9n8kMypxjGUnRtRJvZ"), [98, 97, 114, 102, 111, 111])}`,
 right: `{(["shared"], PeerId("12D3KooWRW9MXWPG9VaFFbyPWxCfZw4geE9n8kMypxjGUnRtRJvZ"), [98, 97, 114, 102, 111, 111])}`', tests/pubsub.rs:124:9
```
Is this a regression or a change in libp2p?

Is this the change in libp2p? <https://github.com/libp2p/rust-libp2p/commit/75ae7b0461b695201cd0161d7a7b7840ff879f7e>
@mxinden

mxinden:
> Is this a regression or a change in libp2p?
> 
> Is this the change in libp2p? [libp2p/rust-libp2p@75ae7b0](https://github.com/libp2p/rust-libp2p/commit/75ae7b0461b695201cd0161d7a7b7840ff879f7e) @mxinden
> 
> Looks like Floodsub topic publishes are not sent back to the local machine anymore?

I am a bit confused, but also not particularly familiar with the Floodsub implementation.

Does rust-ipfs depend on self-published messages to be reported as received messages? I am not sure how https://github.com/libp2p/rust-libp2p/commit/75ae7b0461b695201cd0161d7a7b7840ff879f7e affected that behaviour. Can you expand?

rand0m-cloud:
Oh sorry, I should have given you a bit more context. The test `publish_between_two_nodes` has the following comment that tipped me off:
```rust
    // the order is not defined, but both should see the other's message and the message they sent
    let expected = [
        (&[topic.clone()], &nodes[0].id, b"foobar"),
        (&[topic.clone()], &nodes[1].id, b"barfoo"),
    ]

```
The code expects the `pubsub_subscribe(topic)` to give out two events. One for the message they sent to the other node and the message they originally sent out. I believe that commit I linked changed the behavior to only send out messages to target peers, which would remove the loopback peer. The test asserts that the two events from each node are equivalent to the expected.

The test needs to be updated where each node only expects to hear the message from the other node.

rand0m-cloud:
Ah, I've figured it out. That <https://github.com/libp2p/rust-libp2p/commit/75ae7b0461b695201cd0161d7a7b7840ff879f7e> change highlighted the fact that we didn't call `{add,remove}_node_from_partial_view` on Floodsub. The Pubsub struct maintains its own Peer list and that commit broke our code because we didn't tell Floodsub about the nodes we are interested in. My commit adds a `add_node_from_partial_view` call when we hear a Subscribed event and a `remove_node_from_partial_view` when we remove a peer from Pubsub's list.

All tests pass now!! Moving this PR from a draft into a PR for review. I will do a quick writeup of changes done to the testing suite when I get a chance today.

rand0m-cloud:
Changes to testing:
* `swarm_api`: Add `biased` to tokio::select for consistent results
* `wrong_peerid`: Updated assert statement to account for error.to_string() changes
* `publish_between_two_nodes`: Altered test to use a topic-A and topic-B and nodes only checks for messages sent from a different peer id

koivunej (approved):

koivunej:
Thanks for your work @rand0m-cloud and apologies for the delay once more. I'll take care of the windows build failure.

tests/pubsub.rs
```diff
@@ -54,25 +55,30 @@ async fn publish_between_two_nodes() {
...
-    let topic = "shared".to_owned();
+    let topic_a = "shared-a".to_owned();
+    let topic_b = "shared-b".to_owned();
 
-    let mut a_msgs = nodes[0].pubsub_subscribe(topic.clone()).await.unwrap();
-    let mut b_msgs = nodes[1].pubsub_subscribe(topic.clone()).await.unwrap();
+    // Node A subscribes to Topic B
+    // Node B subscribes to Topic A
+    let mut a_msgs = node_a.pubsub_subscribe(topic_b.clone()).await.unwrap();
+    let mut b_msgs = node_b.pubsub_subscribe(topic_a.clone()).await.unwrap();
```
koivunej, line 67: Unclear why this was done. Is it now impossible to have the two peers in the same topic and see each other in the peers...? I think this should be investigated. I'll check to see if this was made before the partial_view change, or after it.

rand0m-cloud: I mainly changed the test to confirm my hypothesis about which messages were actually being received. At the time, I didn't know which node had what PeerId. After altering the test, I discovered the only messages received were from the localhost.

I'm not opposed to reverting the commit. We would still want the timeout on the `.take`.

koivunej: I think the failed conformance tests could be related to this. Except that they only failed on windows which I don't have around for development. I'll look around, and rerun the windows build.

koivunej: Sadly no luck with the re-run, however the failing tests did change: https://github.com/rs-ipfs/rust-ipfs/runs/5754868818?check_suite_focus=true#step:18:6049 -- need to continue with this tomorrow.

koivunej: Been looking at this a bit. I think you had a good idea with the test case, so it should be kept next to the old one. I cannot really reproduce the failure on the old version, but did some simplifications (from hashset -> vec) and expect all messages (even the echoed locally sent ones). Cannot fathom why hashset was used there originally, but I recall it was added my be.

koivunej:
I would rather not delay this further and a lot of work has been put into this already. Thanks @rand0m-cloud! I ended up ignoring the problematic windows tests pending a) packet capture b) additional logging in the js test. Also added a FIXME for my concern.

bors r+

PR #500, opened 2022-03-17
feat(test): added additional context on why test could fail

rand0m-cloud:
I just spent some time debugging test failures as I looked back through previous versions of rust-ipfs. The test `pre_configured_listening_addrs` assumes that port 4001 is available for binding and hopes that the multiaddr `/ip4/127.0.0.1/tcp/4001` is found in `ipfs.identify()`. I always run an IPFS daemon in the background for web app development and found the test passing when I killed the daemon.

This PR adds additional context to the error message, `is port 4001 available to listen on?`.


koivunej (approved):

koivunej:
This seems straightforward one, thanks for this! As for this particular test, I think the port number could had been anything else, I guess except for 0.

rand0m-cloud:
Yeah, I was facing that and that's why I was working to update the dependency. Now that I think about it, it might have been easier to make a patch level upgrade to build on stable Rust.

koivunej:
Yeah seems like everyone else should had been hitting this as well, did not check if there's a patch version for this.

rand0m-cloud:
FWIW, I've made a [v0.39.2 tag on my repo](https://github.com/libp2p/rust-libp2p/compare/v0.39.1...rand0m-cloud:v0.39.2?expand=1) that fixes the build error in libp2p-noise. I can't seem to open a PR comparing it to v0.39.1 on libp2p. I'm not sure it would be worth my time to make a PR on libp2p and rust-ipfs when my other PR is almost ready.

PR #502, opened 2022-04-05
feat(http): add default values for --bits and --profile in ipfs init

rand0m-cloud:
This PR adds default values for the `ipfs init` command to be more like the go-ipfs `ipfs init`. `--bits 2048` and `--profile default` are the default flags if they not overridden.

I've also included an profiles length check that is more accurate and injects the default profile if none was provided. Plus a small update to an error print statement because 1024 bits for a RSA key is too low.

### DIFF
--- a/http/src/main.rs
+++ b/http/src/main.rs
@@ -15,13 +15,13 @@ enum Options {
     /// with two arguments by default, `--bits 1024` and `--profile test`.
     Init {
         /// Generated key length
-        #[structopt(long)]
+        #[structopt(long, default_value = "2048")]
         bits: NonZeroU16,
         /// List of configuration profiles to apply. Currently only the `Test` and `Default`
         /// profiles are supported.
         ///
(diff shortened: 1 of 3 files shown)


http/src/main.rs
```diff
@@ -95,7 +95,7 @@ fn main() {
                     std::process::exit(1);
                 }
                 Err(config::InitializationError::InvalidRsaKeyLength(bits)) => {
-                    eprintln!("Error: --bits out of range [1024, 16384]: {}", bits);
+                    eprintln!("Error: --bits out of range [2048, 16384]: {}", bits);
```
koivunej, line 98: Good spotting!

koivunej:
Thanks for this @rand0m-cloud. In near-term it would be nice to get rid of the RSA keys. I think I saw there being some support for ipfs' or libp2p compatible ec key serialization in rust-libp2p as well, which would allow removal of lot of complications.

bors r+

PR #507, opened 2022-06-16
refactor: Replace floodsub with gossipsub

dariusc93:
This PR replaces floodsub with gossipsub. Left it as a draft as a starting point to get a bit more insight as to if this is the right starting point to proceed from there. Will be going back after to comment on everything as well that needs commenting or if comments needs adjustments to reflect changes.

~The only issue right now, which I may just be looking over (due to my lack of time in debugging it) is that while I am able to subscribe to the topic, no messages are being published to it nor is the stream receiving anything in the topic.~

EDIT: When the peers are connected (either directly or is discovered via mdns) they are able to see each other.

This [Draft] PR supersede #489

- [x] **New** code is “linted” i.e. code formatting via rustfmt and language idioms via clippy
- [x] There are no extraneous changes like formatting, line reordering, etc. Keep the patch sizes small!

### DIFF
--- a/src/p2p/pubsub.rs
+++ b/src/p2p/pubsub.rs
@@ -17,7 +17,7 @@ use libp2p::core::{
 
 use libp2p::gossipsub::{
     self, Gossipsub, GossipsubEvent, GossipsubMessage, IdentTopic as Topic, MessageAuthenticity,
-    MessageId, TopicHash, ValidationMode,
+    MessageId, TopicHash,
 };
 use libp2p::swarm::{
     ConnectionHandler, DialError, NetworkBehaviour, NetworkBehaviourAction, PollParameters,
@@ -141,11 +141,10 @@ impl Pubsub {
     /// top of the gossip.
     pub fn new(keypair: Keypair) -> Result<Self, Error> {
         let (tx, rx) = channel::unbounded();
-        let config = {
-            gossipsub::GossipsubConfigBuilder::default()
-                .build()
-                .map_err(|e| anyhow::anyhow!("{}", e))?
-        };
+        let config = gossipsub::GossipsubConfigBuilder::default()
+            .build()
+            .map_err(|e| anyhow::anyhow!("{}", e))?;
+
         Ok(Pubsub {
             streams: HashMap::new(),
             peers: HashMap::new(),
@@ -179,7 +178,7 @@ impl Pubsub {
                     }
                     Ok(false) => None,
                     Err(e) => {
-                        debug!("{}", e); //"subscribing to a unsubscribed topic should have succeeded"
+                        error!("Error subscribing to topic: {}", e); //"subscribing to a unsubscribed topic should have succeeded"
                         None
                     }
                 }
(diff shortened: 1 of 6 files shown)


dariusc93:
Should we go on and pass through the gossip result? Eg `Gossipsub::publish` returns `Result<MessageId, PublishError>`. Should `Pubsub::publish` and `IpfsEvent::PubsubPublish` also use the same return? Although the test is failing on a single test (which I will have to dig into soon or probably just change the test although I shouldnt need to change anything, will find out here shortly), testing it code wise it works just fine once the peers are connected to eachother.

src/p2p/pubsub.rs
```diff
@@ -182,22 +192,22 @@ impl Pubsub {
...
     /// Returns true if an existing subscription was dropped, false otherwise
-    pub fn unsubscribe(&mut self, topic: impl Into<String>) -> bool {
-        let topic = Topic::new(topic);
-        if self.streams.remove(&topic).is_some() {
-            assert!(
-                self.floodsub.unsubscribe(topic),
-                "sender removed but unsubscription failed"
-            );
-            true
+    pub fn unsubscribe(&mut self, topic: impl Into<String>) -> Result<bool, Error> {
```
dariusc93, line 195: Used `anyhow::Error` since i was unsure if I should passthrough `Gossipsub::unsubscribe` result, but even if I did it would still leave the matter of what to do for the other part of the if/else statement which is why I used `anyhow::Error`.

src/p2p/pubsub.rs
```diff
@@ -206,10 +216,17 @@ impl Pubsub {
     }
 
     /// Returns the peers known to subscribe to the given topic
-    pub fn subscribed_peers(&self, topic: &Topic) -> Vec<PeerId> {
+    pub fn subscribed_peers(&self, topic: &str) -> Vec<PeerId> {
```
dariusc93, line 219: Not sure but maybe we could use `Gossipsub::mesh_peers` for this in some way since the protocol does keep a mesh peers for a specific topic hash.

tests/pubsub.rs
```diff
@@ -39,13 +39,14 @@ async fn unsubscribe_via_drop() {
...
-#[tokio::test]
-async fn can_publish_without_subscribing() {
-    let a = Node::new("test_node").await;
-    a.pubsub_publish("topic".into(), b"foobar".to_vec())
-        .await
-        .unwrap()
-}
+// This test may not be needed since rust-libp2p gossip implementation doesnt allow publishing without first subscribing(?)
+// #[tokio::test]
+// async fn cant_publish_without_subscribing() {
```
dariusc93, line 44: Commented this out since gossip doesnt support publishing without subscribing. A topic would need to be subscribed and would also need another peer available in the topic for function to succeed.

src/p2p/pubsub.rs
```diff
@@ -132,47 +137,52 @@ impl FusedStream for SubscriptionStream {
...
+        let config = {
+            gossipsub::GossipsubConfigBuilder::default()
+                .build()
+                .map_err(|e| anyhow::anyhow!("{}", e))?
+        };
+        Ok(Pubsub {
             streams: HashMap::new(),
             peers: HashMap::new(),
-            floodsub: Floodsub::from_config(config),
+            gossipsub: Gossipsub::new(MessageAuthenticity::Signed(keypair), config)
```
dariusc93, line 152: Messages should be signed but maybe give some options to allow configuring gossipsub and its authority and maybe expand `GossipsubConfigBuilder` in some way to allow additional configuration?

PR #510, opened 2022-08-19
IPLD fuzzing

evanrichter:
This PR adds a fuzz harness for IPLD decoding, so that the decoder can be more resilient from panics, etc.

        - wanted? 
    - not necessary as it isn't user facing?

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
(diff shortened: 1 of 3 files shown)


koivunej (approved):

koivunej:
Merging regardless.

evanrichter:
awesome, thanks!!

